PR #6, opened 2026-07-23
feat: post-print fan cooldown above 40°C

plastikman:
## What
After a print, keep the blower running until the chamber cools below **40 °C**, then stop — so the chamber self-cools after heating instead of holding residual heat (and a just-tripped over-temp chamber keeps getting cooled).

## Behaviour / decision
Gated on a **RAM-only "heater ran this session" flag** (set when `pb_heater_heat_mode()` goes true), **not** on chamber temperature alone:
- ✅ Fan **continues** cooling after the target drops to 0, while chamber > 40 °C.
- ✅ Never **auto-starts** on temperature — a warm chamber for other reasons (hot day, ambient) does nothing.
- ✅ **Power-cycle-while-hot → fan stays off** (flag isn't persisted). This was the explicit requirement: "no one wants the fan blowing if they didn't just finish a print."
- Disarms once cooled below 40 °C (no re-trigger until the next heat cycle); a NAN/failed chamber read disarms rather than fanning blindly.

## Implementation
All in `pb_policy_tick()` (`components/pb_policy/pb_policy.c`) — the existing fan-control point. Adds `s_heated_this_session` + a `PB_COOLDOWN_TEMP_C` (40 °C) check against the smoothed chamber temp (`pb_ntc_smoothed_c(PB_NTC_CHAMBER)`). No API/CMake changes (`pb_ntc` was already a dependency). Builds clean on ESP-IDF v5.3 (esp32c3).

## Test plan (on device)
1. Set a target ≥ ~45 °C, let the chamber pass 40 °C.
2. Set target 0 (end print) → **fan keeps running**; watch chamber fall; **fan stops when it drops below 40 °C**.
3. Heat > 40 °C, then **power-cycle** → after boot the **fan stays off** despite the hot chamber.
4. Cold boot / warm ambient with no heating → fan stays off.

Credit: cooldown idea from wildtang3nt (@justinh-rahb); session-gating requirement from @plastikman.

### DIFF
--- a/components/pb_policy/pb_policy.c
+++ b/components/pb_policy/pb_policy.c
@@ -50,12 +50,16 @@ void pb_policy_tick(void)
 
     // Post-print cooldown: once the heater has run, keep airflow going after it
     // turns off until the chamber cools below the threshold, then disarm (don't
-    // re-trigger until the next heating session). Smoothed chamber temp; a NAN /
-    // failed reading disarms rather than fanning blindly.
+    // re-trigger until the next heating session). Only trust the smoothed temp if
+    // the latest read was OK — on a sensor fault pb_ntc_smoothed_c() keeps
+    // returning the last (stale) moving average, NOT NaN, which could otherwise
+    // pin the fan on forever. On fault/no-reading, disarm; the heater's own fault
+    // path (pb_heater_is_faulted) still drives airflow if it latches.
     bool cooldown = false;
     if (!heat && s_heated_this_session) {
         float chamber_c = pb_ntc_smoothed_c(PB_NTC_CHAMBER);
-        if (!isnan(chamber_c) && chamber_c > PB_COOLDOWN_TEMP_C) {
+        if (pb_ntc_last_status(PB_NTC_CHAMBER) == PB_NTC_OK
+                && !isnan(chamber_c) && chamber_c > PB_COOLDOWN_TEMP_C) {
             cooldown = true;
         } else {
             s_heated_this_session = false;
--- a/components/pb_policy/pb_policy.c
+++ b/components/pb_policy/pb_policy.c
@@ -2,16 +2,31 @@
 #include "pb_policy.h"
 #include "pb_heater.h"
 #include "pb_fan.h"
+#include "pb_ntc.h"
 
 #include "esp_log.h"
+#include <math.h>
 
 static const char *TAG = "pb_policy";
 
+// Post-print cooldown: after the heater has run this session, keep the fan going
+// until the chamber falls below this threshold. NOT a temperature trigger on its
+// own — it is gated by s_heated_this_session (see below).
+#define PB_COOLDOWN_TEMP_C 40.0f
+
 static uint8_t s_requested_fan;
+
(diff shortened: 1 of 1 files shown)


plastikman:
✅ **Validated on hardware** (Panda on USB, driven via the U1 full stack):

- Set target 45 °C → heated; at **43.3 °C** cut the target to 0.
- **Fan kept running** through the cooldown (43 → 40 °C) with the heater off — the new behaviour.
- **Fan turned off below 40 °C** (brief toggle right at the 40 °C threshold as residual heat bleeds back — harmless; a small hysteresis would smooth it if desired).

Matches the spec (continue-after-print, never temp-trigger). Ready to merge.

plastikman:
Fixed in the latest commit: cooldown now checks `pb_ntc_last_status(PB_NTC_CHAMBER) == PB_NTC_OK` before trusting the smoothed temp, so a chamber-sensor fault (which leaves `pb_ntc_smoothed_c()` at a stale value, not NaN) **disarms** the cooldown instead of pinning the fan on. If the fault latches the heater, `pb_heater_is_faulted()` still drives airflow. Builds clean on ESP-IDF v5.3. Thanks for the catch.

components/pb_policy/pb_policy.c
```diff
@@ -28,11 +43,30 @@ void pb_policy_tick(void)
...
+
+    // Post-print cooldown: once the heater has run, keep airflow going after it
+    // turns off until the chamber cools below the threshold, then disarm (don't
+    // re-trigger until the next heating session). Smoothed chamber temp; a NAN /
+    // failed reading disarms rather than fanning blindly.
+    bool cooldown = false;
+    if (!heat && s_heated_this_session) {
+        float chamber_c = pb_ntc_smoothed_c(PB_NTC_CHAMBER);
+        if (!isnan(chamber_c) && chamber_c > PB_COOLDOWN_TEMP_C) {
+            cooldown = true;
```
justinh-rahb, line 59: ### [P2] Stale average masks sensor failure
`pb_ntc_smoothed_c()` retains its last finite moving average when `pb_ntc_read()` fails, so `isnan(chamber_c)` does not detect a thermistor fault. During cooldown, a cached value above 40 °C can keep the fan running indefinitely rather than disarming as documented. Check `pb_ntc_last_status(PB_NTC_CHAMBER) == PB_NTC_OK` before using the smoothed value.

justinh-rahb (commented):
- **[P2] Sensor failure does not actually disarm cooldown.** `pb_ntc_smoothed_c()` retains and returns its previous moving average after `pb_ntc_read()` reports a sensor fault. If the chamber thermistor disconnects during cooldown while the cached value is above 40 °C, the fan may run indefinitely instead of disarming as the PR claims. Check `pb_ntc_last_status(PB_NTC_CHAMBER) == PB_NTC_OK` before accepting the smoothed value.

The normal cooldown behavior is otherwise coherent and hardware-validated. PR head `12faec1` builds cleanly under ESP-IDF 5.3.1, and GitHub CI passes. I’d request this small correction before merging.

PR #10, opened 2026-07-23
feat(phase-a): configurable safety settings + real status LEDs

plastikman:
Phase A of the [iteration-2 plan](plans/iteration-2-stock-parity-and-config.md) (stock parity + config). Adds runtime-tunable heater safety settings persisted to NVS, an HTTP + web-UI surface for them, and a real indicator-LED driver — with `pb_heater` no longer touching any GPIO LED.

## What's in it

**A1 — `pb_heater` runtime settings (sole SSR owner)**
- Settable **max-target ceiling** (default 70 °C) and **comms-watchdog timeout** (default 5 min), both guarded by the existing `s_mux` and persisted to NVS (`app_nvs`: `heat_max_c` centi-°C, `heat_comms_ms`).
- `pb_heater_load_config()` reads + clamps on boot (after `nvs_init()`); setters clamp, persist, and pull a live target down if it now exceeds the cap.
- **Safety invariant preserved:** no NVS/API path can heat past 70 °C; the comms deadman can be *shortened* but never disabled or extended beyond 5 min. The fixed **85 °C chamber / 105 °C PTC** cutoffs are untouched and non-configurable.

**A2 — HTTP**
- `GET /settings` (bounds + current values) and auth-gated `POST /settings` (fields `max`, `comms_ms`, each optional; echoes the *clamped* effective values).
- `/status` now reports live `max`, `max_abs`, and `comms_ms`. `max_uri_handlers` 12→14.

**A3 — Portal**
- An **Advanced / Safety** card on the status dashboard, driven client-side from `/settings` (own `fetch`, inline "Saved ✓", no reboot). The set-temp input ceiling now tracks the configured `max` instead of a hardcoded 70.

**A4 — new `pb_leds` component**
- 3-LED pattern driver (OFF / SOLID / BLINK / BLINK_SLOW / CODE) on a 50 ms task, atomic per-LED state. `pb_policy` drives **K1**: fault → BLINK, heat → SOLID, else OFF — a latched fault is now visible locally instead of going dark. The LED drive was removed from `pb_heater.c` so it owns no LED.

## Deviations from the plan (both simplifications)
1. The Advanced card lives on the **status dashboard** and is driven **client-side** from `/settings`, so `pb_portal` needs **no** `pb_heater` include / CMake dep.
2. `GET /settings` was added alongside `POST` (bounds + values in one read).

## Testing
- `idf.py build` clean.
- **On-device (pending, USB reflash):** K1 solid on heat / blink on fault / off idle; `GET /settings` shows bounds; POST `max=60` then a target of 65 clamps to 60; setting persists across reboot; `max=90` clamps to 70; a shortened `comms_ms` latches off sooner; portal Advanced card round-trips.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,10 @@ jobs:
           esp_idf_version: v5.3.5
           target: esp32c3
           path: '.'
+          # Release builds layer sdkconfig.release on top of the defaults to enable
+          # CONFIG_PB_POWER_LED (drives the front-panel Power LED on GPIO21, giving up
+          # console TX). Dev builds (plain `idf.py build`) omit it and keep the console.
+          command: idf.py -DSDKCONFIG_DEFAULTS="sdkconfig.defaults;sdkconfig.release" build
 
       - name: Package release artifacts
         env:
--- a/components/pb_board/include/pb_board.h
+++ b/components/pb_board/include/pb_board.h
@@ -36,10 +36,15 @@
 // the thermistor resistance (see pb_ntc). level 0 -> 82 kOhm, level 1 -> 33 kOhm.
 #define PB_GPIO_RREF_STRAP   GPIO_NUM_19
 
-// -------- Button LEDs (active-high via 1K) ----------------------------------
-#define PB_GPIO_LED_K1       GPIO_NUM_6
-#define PB_GPIO_LED_K2       GPIO_NUM_5
-#define PB_GPIO_LED_K3       GPIO_NUM_4
+// -------- Front-panel LEDs (active-high, direct push-pull; NOT a matrix) -----
+// Confirmed by stock-firmware RE (byte-identical across stock 1.0.1-1.0.4).
+// K1/K2/K3 are the three mode LEDs; the "Power" LED is on GPIO21 — which is also
+// the UART0 console TX pin, so it is only driven when CONFIG_PB_POWER_LED is set
+// (release builds); otherwise GPIO21 stays the serial console. See pb_leds.
+#define PB_GPIO_LED_K1       GPIO_NUM_6    // "Auto"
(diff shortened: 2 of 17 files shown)


justinh-rahb (approved):
Reviewed the current head (233559e) against main and rebuilt it locally with ESP-IDF 5.3 (Python 3.10.12): clean build, 47% app-partition headroom.

No merge-blocking findings. The safety envelope remains fail-closed: the configurable ceiling cannot exceed 70 °C, the comms deadman cannot be disabled or extended beyond five minutes, persisted values are clamped again on load, lowering the ceiling pulls down a live target, and the fixed 85 °C chamber / 105 °C PTC trips are untouched. NVS is loaded only after initialization while the boot target remains zero. The HTTP mutation is auth-gated, and LED ownership/mapping now correctly puts heat/fault indication on the physical On LED (K2/GPIO5), with the mapping bench-validated.

Still worth completing the documented hardware pass for settings persistence, live target clamping, shortened watchdog timing, and portal round-trip before release.

Non-blocking robustness follow-up: the settings setters currently ignore nvs_open/nvs_set_u32/nvs_commit failures, so /settings may report “Saved” even when the change is runtime-only and will revert after reboot. Propagating persistence errors would make that status truthful.

plastikman:
Added a follow-up commit (64c65ce): **Power LED (GPIO21) as the heating indicator**, gated behind `CONFIG_PB_POWER_LED` (release-only, since GPIO21 is the console TX pin). Stock-firmware RE confirmed the 4 panel LEDs are direct active-high outputs (no matrix); GPIO21 being the console TX is why Power was stuck on. Validated on hardware: Power off at idle, Power+On solid while heating (matches stock); dev build keeps the console. This rounds out Phase A's "real status LEDs."

PR #16, opened 2026-07-24
feat(policy): persist remembered mode parameters

justinh-rahb:
Second of three stacked PRs closing out Phase C. **Stacked on #15** - review that
one first; this diff is against it.

No button code here either. This closes a gap found while planning Phase C.

## Problem

`plans/iteration-2-stock-parity-and-config.md` marks Phase B complete including
"**B5** boot-OFF, params-only NVS persistence". Boot-OFF shipped. The persistence
did not - `pb_policy.c` had no NVS code and none of the documented `md_*` keys
existed.

Two consequences: the dashboard pre-filled hardcoded values rather than the
device's remembered values, and the front-panel buttons in PR 3 had nowhere to
read "the target the user last chose" from.

## Change

Remember the last accepted values per mode and expose them as `params` in
`GET /api/v2/state`:

| Key | Meaning | Default | Clamp |
|---|---|---|---|
| `md_last` | manual POWER_ON target | 50 C | 30-70 C |
| `md_auto_tgt` | AUTO target | 60 C | 30-70 C |
| `md_auto_bed` | AUTO bed threshold | 100 C | 40-120 C |
| `md_dry_tgt` | drying target | 60 C | 30-70 C |
| `md_dry_hrs` | drying duration | 12 h | 1-12 h |

**Writes go through one dedicated worker task**, not inline under `s_lock`.
Writing outside the lock is right for latency, but it would let concurrent HTTP
and button commands commit out of order and strand an older preset on flash.
With exactly one writer that always commits the latest canonical snapshot, that
hazard cannot occur. The worker records the **post-clamp** value the device
actually applied (never the raw request), skips unchanged keys to spare the
flash, commits explicitly, and leaves the state dirty on failure so the next
accepted command retries.

The remembered parameters are part of `pb_policy_snapshot_t`, so one state
response cannot mix policy fields from different revisions. The dashboard
hydrates all five controls from `state.params` while preserving a value the user
is actively editing.

## Safety

`pb_policy_load_params()` touches no mode, target, deadline, or lease - it is
parameters only, so **the device still boots OFF**. There is a test asserting
exactly that, including that a drying session active before a reboot does not
come back.

Missing values retain their defaults. Finite stored temperatures are clamped to
their safe envelopes, non-finite temperatures fall back to defaults
defensively, and the stored duration is range-checked.

## Verification

- [x] `tests/run_policy_host_test.sh` - fake in-memory NVS coverage for empty
      flash defaults, load clamping, only-changed-key writes, canonical
      post-clamp persistence, failed-write retry, coherent state snapshots, and
      params-survive-reboot-but-mode-does-not.
- [x] `tests/check_api_v2_contract.sh` - pins the `params` fields, coherent
      snapshot serialization, and dashboard consumption.
- [x] `tests/check_dashboard_js.mjs`.
- [x] ESP-IDF 5.3.1 release build (`sdkconfig.release`) at `e0f405c`.
- [x] ESP-IDF 5.3.1 devboard HIL build (`sdkconfig.hil-devboard`) and
      `tests/check_hil_compileout.sh`.
- [x] Real Panda bench at `e0f405c`: authenticated commands recorded manual
      30 C, AUTO 31 C / 119 C bed threshold, and drying 30 C / 7 h. Every
      observed command response kept `heater.output=false`; the unit was
      returned OFF after each mode.
- [x] On-device hard reset through the USB serial bridge: `boot_id` changed,
      revision reset to 1, source/mode returned as `boot`/`off`, requested and
      effective targets were 0, heater demand/output and lease were false, and
      all five distinct parameters returned unchanged from NVS.


justinh-rahb:
Superseded by a fresh PR after the squash-merge of #15 required rebasing this branch onto `main` — reopening was blocked because the head branch was force-pushed while closed. Same reviewed commits (`ef1092d` + `e0f405c`), now rebased. See the new PR.

PR #20, opened 2026-07-24
fix(policy): tighten button feedback and target bounds

justinh-rahb:
## Summary

- Wake the control task after every accepted front-panel state change, so mode
  LEDs and outputs update promptly instead of waiting up to 500 ms for the
  periodic tick.
- Preserve the single-SSR-writer invariant: button actions still only notify
  the control task, which runs the complete policy/safety tick.
- Log rejected mode presses with the actual policy result instead of reporting
  a misleading transition to `off`.
- Clamp remembered manual, Auto, and Dry targets to the live
  runtime-configured heater maximum when loading them from NVS.
- Document that the panic-off latch is RAM-only and clears on reboot while the
  device still boots fully OFF.

## Root cause

The wake callback was used only by panic-off, and the short-press handler
discarded mode-setter results. Separately, boot-time parameter loading used the
absolute 70 °C ceiling instead of the already-loaded runtime ceiling.

## Validation

- `sh tests/run_policy_host_test.sh`
- `sh tests/run_buttons_host_test.sh`
- `sh tests/check_api_v2_contract.sh`
- `node tests/check_dashboard_js.mjs`
- `python3 -m unittest tests/test_hil_runner.py` — 11 tests
- ESP-IDF v5.3 production build
- ESP-IDF v5.3 USB HIL build + compile-out check
- ESP-IDF v5.3 UART HIL build + compile-out check

New host coverage verifies accepted button wakes, no wake for an already-off
no-op or fault rejection, accurate fault-rejection logging, and
runtime-ceiling clamping of all remembered heat targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
+### Fixed
+- **Prompt control feedback across every command source.** Accepted front-panel,
+  Web UI, Klipper, and HIL mode commands now wake the full control task
+  immediately instead of waiting up to 500 ms for the periodic tick, so panel
+  LEDs and outputs track authoritative state promptly. Rejected button commands
+  now log the actual policy result, and remembered mode targets are clamped to
+  the runtime-configured heater maximum as soon as they load from NVS.
+
 ## [0.4.0] - 2026-07-24
 
 Phase C — the physical front panel comes alive. All four buttons and all four
--- a/components/pb_policy/include/pb_policy.h
+++ b/components/pb_policy/include/pb_policy.h
@@ -55,7 +55,8 @@ typedef struct {
 // caller does not supply its own (front-panel buttons) and the values the UI
 // pre-fills from.  These are the ONLY policy state that survives a reboot --
 // never a mode, target, deadline, heating flag, or lease.  See
-// pb_policy_load_params().
+// pb_policy_load_params(). Target parameters are clamped to the heater's live
+// runtime-configured maximum when loaded.
 typedef struct {
     float manual_target_c;        // last accepted POWER_ON target
     float auto_target_c;          // last accepted AUTO target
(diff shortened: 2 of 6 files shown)


plastikman:
## Review — ✅ approve

_AI-assisted review (Claude Code), read-only; verified against merged main + the PR branch._

This cleanly resolves four of the follow-ups from the #17/#18 post-merge audits, and does it carefully. No blockers.

**1. Wake on every accepted command (was: LED lag up to ~500 ms).** `wake_control_task()` is now called from `set_power_on`/`set_auto`/`start_drying`/`set_mode_off`/`clear_fault`/panic-off — and crucially only on the **accepted** path (rejections return before the give+wake), so a fault-rejected press does **not** wake. Nice bonus: this also makes Web/Klipper/HIL commands update outputs promptly, not just buttons. **Single-SSR-writer invariant preserved** — the wake is just a task notification; the control task still runs the full `pb_policy_tick()` (no partial "drop the SSR" path), and its absolute-deadline/one-extra-tick/watchdog-fed-after-tick guardrails mean extra notifications can't drag or accelerate the safety schedule.
- Verified the **Power already-off no-op** logs only (`"btn: power (already off)"`) and does **not** call `set_mode_off` → no spurious wake/revision churn. ✅

**2. Accurate rejection logging.** `button_toggle_mode` now returns `pb_policy_result_t`, and `on_button` logs `"btn: <x> rejected (<reason>)"` via `pb_policy_result_str()` (confirmed it exists) instead of the old misleading `"-> off"`. ✅

**3. Runtime-max clamp on NVS load.** `params_clamp` now clamps manual/auto/dry targets to `pb_heater_get_max_target_c()` instead of the absolute 70 °C. **Ordering verified** — `app_main` calls `pb_heater_load_config()` (line 275) before `pb_policy_load_params()` (line 276), so the live configured ceiling is available when params load; the code comment is correct. Bonus: the `clamp_or_default` tweak (`v = dflt` then clamp, instead of early-return) means even a default above a user-lowered max is now clamped down — a strict improvement. ✅

**4. SAFETY.md** now states the panic-off latch is RAM-only / clears on reboot (boots fully OFF). ✅

**Tests:** new host coverage for accepted-wake, no-wake on no-op/rejection, accurate rejection logging, and runtime-ceiling clamp; validation ran the full policy/buttons/contract/dashboard/HIL suites + production and both HIL builds.

Good to merge.

plastikman (approved):

PR #24, opened 2026-07-24
feat(safety): B2 — persistent fault latch + hardened residual-heat purge

plastikman:
Implements **B2** from the iteration-2 roadmap — the deferred safety work — as an independently-reviewable PR. Two features:

## 1. Persistent fault latch (`pb_heater`)
Previously a latched safety fault was RAM-only, so a power cycle cleared it and the device could come back up ready to heat.

- A **hazard-driven** trip (PTC/chamber over-temp, sensor fault while heating, comms-loss watchdog) is now written to NVS on the latching transition and **restored at boot** → device comes up **heater-OFF, commands inhibited** until an explicit clear after the condition recovers.
- Persisted cause is a stable **`u8` enum** (`pb_fault_reason_t`); a live reason string still carries session detail (design decision: enum for the persisted code + RAM string for detail).
- Boot restore is **fail-safe**, via a pure header-inline `pb_heater_fault_decide()`: unreadable store → latched; corrupt code → generic latch; fresh device (no namespace) → not a fault.
- Clearing is **persist-first**: `pb_heater_clear_fault()` clears NVS before RAM and returns `esp_err_t`; a failed persist keeps the latch and surfaces as **HTTP 500** (`PB_POLICY_PERSIST_FAILED`) — never a false "cleared".
- **Not** persisted: the user **panic-off** (documented reboot-clears manual stop) and the permanent **inhibit** (a power cycle must lift it). Active mode/target/deadline/lease still never persist — boot stays OFF.

## 2. Hardened residual-heat purge (`pb_policy`)
The session-gated cooldown fan now has proper **hysteresis** (engage at ≥ 40 °C, release only once **both** chamber and PTC are < 37 °C; an unknown/faulted sensor keeps the fan on) instead of a single chamber-only threshold — factored into a pure `pb_purge_decide()`.
- Still strictly **session-gated**: never starts on temperature alone, and a **power-cycle-while-hot does not spin the fan** (honors the prior explicit decision on this).

## Tests / CI
- New `tests/run_heater_fault_host_test.sh` (fail-safe boot-restore decision) **wired into `firmware-build.yml`** as a required step.
- `pb_policy` host test gains purge-hysteresis and persist-failure-clear cases.
- All host tests + contract/dashboard gates pass locally; ESP-IDF build clean.

## Validation
Bench OTA confirmed a normal device (existing `app_nvs`, no fault keys) boots **un-latched** (`fault_latched=false`, mode off) — no false fail-safe trip. A real over-temp can't be forced safely on the bench, so the fail-safe/persistence decision logic is covered by the host tests.

Docs: `SAFETY.md` (persistent-fault + purge sections) and `CHANGELOG` updated. Ships as **v0.6.0** after review.

### DIFF
--- a/.github/workflows/firmware-build.yml
+++ b/.github/workflows/firmware-build.yml
@@ -51,6 +51,9 @@ jobs:
       - name: Test sensor-calibration safety clamp (set + NVS-load bound)
         run: sh tests/run_ntc_calibration_host_test.sh
 
+      - name: Test persistent fault-latch boot restore (fail-safe)
+        run: sh tests/run_heater_fault_host_test.sh
+
       - name: Test HIL scenario runner
         run: python3 -m unittest tests/test_hil_runner.py
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,30 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
+### Added
+- **Persistent safety-fault latch (B2).** A hazard-driven safety trip (PTC/chamber
+  over-temp, sensor fault while heating, comms-loss watchdog) now survives a power
+  cycle: it is written to NVS on the latching transition and restored at boot, so a
+  device that tripped before losing power comes back up **heater-OFF with commands
+  inhibited** instead of silently ready to heat. NVS is loaded **before** the
+  control task starts, and a failed persist is **retried** (a pending flag keeps
+  trying until the commit lands) so an early or transiently-failed write can't lose
+  the latch across a reboot. The persisted cause is a stable numeric code (a live
+  reason string still carries session detail). Boot restore is **fail-safe** — if
+  the stored state cannot be read reliably the device comes up latched. Clearing a
+  fault is **persist-first**: if the NVS clear fails the latch is kept and the API
+  returns HTTP 500 (`persist_failed`) rather than falsely reporting it cleared.
+  User panic-off and the permanent inhibit are not persisted (documented
+  reboot-clears behavior); active mode/target/deadline/lease still never persist.
+
+### Changed
+- **Residual-heat purge hardened.** The post-heat cooldown fan now uses hysteresis
+  (engage at ≥ 40 °C, release only once **both** the chamber and PTC sensors are
+  below 37 °C) instead of a single chamber-only threshold, and it also runs when a
+  sensor reading is **unknown** right as heating ends (can't confirm cool → fail
+  safe). It remains strictly session-gated: the fan never starts on temperature
+  alone, and a power-cycle-while-hot does **not** spin it.
+
 ## [0.5.2] - 2026-07-24
 
 ### Fixed
--- a/components/pb_heater/include/pb_heater.h
+++ b/components/pb_heater/include/pb_heater.h
@@ -13,6 +13,55 @@
 #include <stdint.h>
 #include "esp_err.h"
 
+// Machine-readable fault cause. Persisted across reboot as a u8 (see
+// pb_heater_load_fault); a live free-form reason string (pb_heater_fault_reason)
+// carries session detail that is NOT persisted. Append new causes at the END so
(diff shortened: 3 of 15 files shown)


plastikman:
Rebased onto `main` (now v0.5.2) and addressed the review feedback in commit `fix(safety): address B2 review`:

**[Blocker] Fault persistence could be lost across reboot**
- **NVS is now loaded before the control task starts** (`app_main`): config, fault latch, calibration, LEDs, and params all load right after `nvs_init()`, *before* `xTaskCreate(control_task)`. So a fault that trips on the first tick — or during early boot — can actually be persisted. NVS init is local + fast; only network bring-up stays after the safety loop.
- **Deferred-persist retry**: `pb_heater` keeps a pending/dirty flag — if a latch persist fails (NVS busy, or an early fault), `pb_heater_tick()` retries (~2 s throttle) until it commits, so a latch that only lived in RAM can no longer be lost on reboot.
- **Race-safe**: latch persist / clear / retry are serialized under a new mutex, and the retry re-reads the latch state *inside* the lock, so a `clear_fault` that lands during a pending retry is never overwritten with a stale "latched".

**[Minor] Purge start edge**
- `pb_purge_decide` now starts the cooldown purge unless **both** sensors are known and below the latch temp — so a sensor that's unknown right as heating ends starts the fan (can't confirm cool → fail safe) instead of skipping the purge. Added a host-test case.

All host tests + contract gate + ESP-IDF build pass; bench boots clean (no false latch) after the reorder. Per the plan this still ships as **v0.6.0** after isolated hardware validation of the persist-across-reboot + fault-recovery paths (the fail-safe/decide logic is host-tested; a real over-temp can't be forced safely on the bench, so that path is best exercised via the HIL dev-board's sensor-fault injection).

justinh-rahb (approved):
Hardware qualification complete on exact head `1481ad7a782450f37ab2201c317d37e1ccbf846c` (ESP-IDF 5.3.1); CI is also green on this SHA.

**Devboard UART HIL (GPIO-safe target)**
- Full suite PASS: AUTO environment/fail-off, front-panel buttons, ownership/leases, chamber sensor-open trip, PTC over-temp trip, forced fault airflow, explicit clear, zero-cross injection.
- Compile-out audit PASS (`mains_gpio_compiled_out=true`; Panda relay/fan/ADC backends absent).
- Added lab-only reboot qualification in a disposable checkout:
  1. armed logical heat and injected chamber-open;
  2. confirmed hazard latch OFF + fan 100%;
  3. hard-reset without erasing NVS;
  4. confirmed latch restored as `source=safety`, mode/output OFF, reason `chamber sensor fault`, fan 100%;
  5. explicitly cleared, reset again, and confirmed it remained cleared/OFF.
- Devboard HIL image SHA-256: `3bde2e4eb85b539de22f4c84485b05ec98d8ef0e8a406b84739a31ef2137a1e9`.

**Real Panda UART HIL**
- Non-heating smoke PASS: real target identified, both NTCs valid, live zero-cross, unconditional OFF.
- Supervised 40 °C heater test PASS: real SSR + fan engaged; PTC rose 34.2→39.9 °C; OFF dropped demand/output immediately.
- Residual purge latched at PTC 40.6 °C, remained fan 100% through the hysteresis band, then released only after both sensors were below 37 °C (final chamber 33.2 °C / PTC 36.0 °C, fan 0, no fault).
- Panda HIL image SHA-256: `fc1a61ad730c9d209c004492d397f32267d283b0b785c0564194adb4e6e1c096`.

The persistence blocker and purge-start edge are resolved and validated. Approved.

**Non-blocking HIL tooling follow-up:** `tools/hil.py` currently closes serial without an unconditional Panda OFF if a heater scenario aborts. For the supervised test I used a lab-only `finally` guard that attempted `off` before closing. We should upstream that guard before committing real-heater scenarios to the suite; this does not affect runtime firmware behavior or block this PR.

PR #28, opened 2026-07-24
Phase E: CI static-analysis gate + host/sim fault coverage (v0.7.0)

plastikman:
Phase E — the "final" roadmap punchlist item: CI/static-analysis hardening + broader host/simulation fault coverage. Targets **v0.7.0**. **No firmware runtime behavior change** — tests, CI, docs, and two behavior-preserving pure-helper extractions.

## E2 — host/simulation fault coverage
- Extracted `pb_heater_eval_trip()` (pure header inline) encoding the exact safety-trip priority ladder; `pb_heater_tick()` now calls it (verified behavior-identical: same order, same `*_ok` gating, same NAN-masking protection). New `pb_heater_safety_host_test` covers over-temp priority (PTC outranks chamber, fires regardless of armed), fail-closed on non-OK/open/short/uninit sensor while armed, NAN never masking a trip, and the armed-only comms watchdog.
- Extracted `pb_ntc_classify()` (pure header inline) for the open/short/rail fail-status mapping; `pb_ntc_read()` now calls it (behavior-identical — the only change is computing `v` a hair earlier, side-effect-free). New `pb_ntc_status_host_test` covers OPEN/SHORT/OK boundaries + the rail guard.

## E1 — CI / static-analysis gate (`firmware-build.yml`)
- **cppcheck** over `components/` + `main/`, failing on any error/warning/performance/portability finding (one unmodellable false positive suppressed inline; the genuine finding it surfaced — a `%u`/`int` format mismatch in the Moonraker WS URI — fixed with a value-identical cast).
- **Warnings-as-errors** (`-Wall -Wextra -Werror`) applied PRIVATE to every first-party `pb_*` component (never ESP-IDF's own).
- The two new host tests **and** the pre-existing `pb_buttons` host test wired into CI as required steps.

## Not done (documented in the plan)
- Dependency-lock diff gate: N/A (`dependencies.lock` is gitignored/generated here).
- clang-format: skipped (a repo-wide reformat would bury the diff).
- OTA-rollback / corrupt-NVS / reconnect-storm sims: no first-party *pure* logic to extract (delegated to ESP-IDF / covered by existing policy + contract gates).
- Native-USB runtime qualification + real-Panda physical HIL sign-off: hardware steps.

## Validation (all local + bench)
Default + release + both HIL builds (with `-Werror`) PASS; all 6 host tests + `test_hil_runner` (11) + `check_api_v2_contract.sh` PASS; **cppcheck exit 0**. Bench OTA: sensors read (`status=ok`), not faulted, Moonraker connects — refactored sensor/trip paths behavior-identical on hardware.

### DIFF
--- a/.github/workflows/firmware-build.yml
+++ b/.github/workflows/firmware-build.yml
@@ -46,20 +46,64 @@ jobs:
       - name: Test authoritative policy state machine
         run: sh tests/run_policy_host_test.sh
 
+      - name: Test front-panel button state machine
+        run: sh tests/run_buttons_host_test.sh
+
       - name: Test sensor-calibration safety clamp (set + NVS-load bound)
         run: sh tests/run_ntc_calibration_host_test.sh
+
+      - name: Test NTC sample classifier (open / short / rail fail-status)
+        run: sh tests/run_ntc_status_host_test.sh
 
       - name: Test persistent fault-latch boot restore (fail-safe)
         run: sh tests/run_heater_fault_host_test.sh
 
+      - name: Test heater safety-trip decision (over-temp / fail-closed / watchdog)
+        run: sh tests/run_heater_safety_host_test.sh
+
       - name: Test HIL scenario runner
         run: python3 -m unittest tests/test_hil_runner.py
+
+      # E1 static analysis: cppcheck over OUR OWN code (components/ + main/). Fails on
+      # any error/warning/performance/portability finding (--error-exitcode). The C
+      # `error` checks are always on. False positives cppcheck's parser cannot model
+      # (e.g. GNU asm() labels on extern binary-blob symbols) are suppressed inline
+      # at the site with a justification; the missing-ESP-IDF-headers noise and the
+      # normal-check-level branch-limit note are suppressed globally (informational).
+      - name: Static analysis (cppcheck)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cppcheck
+          cppcheck --version
+          INCS=""
+          for d in components/*/include; do INCS="$INCS -I$d"; done
+          cppcheck \
+            --enable=warning,performance,portability \
+            --inline-suppr \
+            --suppress=missingInclude \
+            --suppress=missingIncludeSystem \
+            --suppress=normalCheckLevelMaxBranches \
+            --error-exitcode=2 \
+            --std=c11 \
+            $INCS \
+            components/ main/
 
       - name: Check API v2 contract and dashboard ownership
         run: |
           sh tests/check_api_v2_contract.sh
           node tests/check_dashboard_js.mjs
 
+      # E1 warnings-as-errors: enforced inside the build itself — the top-level
+      # CMakeLists applies `-Wall -Wextra -Werror` PRIVATE to every first-party pb_*
+      # component (never to ESP-IDF's own code). The default + HIL builds here and
+      # the release build in release.yml all fail on any warning in our components
+      # (verified clean under default AND sdkconfig.release).
+      #
+      # E1 dependency-lock validation is intentionally NOT a git-diff gate here:
+      # `dependencies.lock` is .gitignored (generated, not version-controlled), so
+      # there is no committed lock to drift. The ESP-IDF build step re-solves and
+      # validates the managed-component dependencies against the pinned manifests
+      # (components/*/idf_component.yml) on every run, which is the validation.
       - name: Build (ESP-IDF v5.3)
         uses: espressif/esp-idf-ci-action@9d38657f3d789ca759b2b37aaf5ceffbc42c4f0d # v1
         with:
--- a/.github/workflows/firmware-build.yml
+++ b/.github/workflows/firmware-build.yml
@@ -118,6 +118,7 @@ jobs:
           target: esp32c3
           path: '.'
           command: |
+            set -e   # a failed idf.py must fail the step, not be masked by the next command
             idf.py -B build-hil-devboard \
               -DSDKCONFIG="$PWD/build-hil-devboard/sdkconfig.profile" \
               -DSDKCONFIG_DEFAULTS="sdkconfig.defaults;sdkconfig.hil-devboard" \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,3 +36,18 @@ message(STATUS "DragonBreath version: ${PROJECT_VER}")
 
 include($ENV{IDF_PATH}/tools/cmake/project.cmake)
 project(dragonbreath)
+
+# E1 (Phase E) — warnings-as-errors for OUR OWN first-party components only.
+# Applied here (after project(), when the per-component library targets exist) as a
+# PRIVATE option so it never leaks to ESP-IDF's own components or to dependents.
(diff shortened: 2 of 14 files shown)


.github/workflows/firmware-build.yml
```diff
@@ -46,20 +46,64 @@ jobs:
...
+      # E1 warnings-as-errors: enforced inside the build itself — the top-level
+      # CMakeLists applies `-Wall -Wextra -Werror` PRIVATE to every first-party pb_*
+      # component (never to ESP-IDF's own code). The default + HIL builds here and
+      # the release build in release.yml all fail on any warning in our components
+      # (verified clean under default AND sdkconfig.release).
+      #
+      # E1 dependency-lock validation is intentionally NOT a git-diff gate here:
+      # `dependencies.lock` is .gitignored (generated, not version-controlled), so
+      # there is no committed lock to drift. The ESP-IDF build step re-solves and
+      # validates the managed-component dependencies against the pinned manifests
```
justinh-rahb, line 105: [P1] These manifests are not actually pinned: `pb_moonraker` specifies `^1.3.0` and `pb_wifi` specifies `^1.4.0`, while `dependencies.lock` is ignored. A clean build currently resolves websocket-client 1.8.0 and mDNS 1.11.3, but the same source commit can silently resolve newer managed-component code later. There is also an IDF split: this workflow's `v5.3` resolved 5.3.0 in CI, while releases use 5.3.5. Please standardize CI/release on the same exact IDF version, commit the generated lockfile, and make CI fail if reconfigure/build changes it. That gives the release provenance claim an actual immutable dependency graph.

justinh-rahb (commented):
Overall this is a good hardening pass. I found no heater-safety regression in the two pure helper extractions, and the expanded host coverage exercises the important trip ordering and sensor classifier boundaries. I verified the exact head with all six C host suites, the 11-case HIL runner, API/dashboard checks, and both default and release ESP-IDF builds. CI is also green. One substantive reproducibility issue and two smaller CI-coverage notes are inline.

CMakeLists.txt
```diff
@@ -36,3 +36,18 @@ message(STATUS "DragonBreath version: ${PROJECT_VER}")
...
+
+# E1 (Phase E) — warnings-as-errors for OUR OWN first-party components only.
+# Applied here (after project(), when the per-component library targets exist) as a
+# PRIVATE option so it never leaks to ESP-IDF's own components or to dependents.
+# These are the pb_* components auto-discovered from components/; ESP-IDF framework
+# code is deliberately excluded (we do not control its warning cleanliness). Both
+# the default and release builds must stay clean under these flags.
+set(DB_OWN_COMPONENTS
+    pb_board pb_buttons pb_evlog pb_fan pb_heater pb_hil pb_httpd
+    pb_leds pb_moonraker pb_ntc pb_policy pb_portal pb_wifi)
```
justinh-rahb: [P2] `main/app_main.c` is first-party code too, but `main` is omitted from this target list. I checked the generated compile database: the `pb_*` commands end with the added `-Werror`, while `main/app_main.c` does not. Adding `main` here would make the warnings-as-errors claim cover all of our compiled C, not only components.

.github/workflows/firmware-build.yml
```diff
@@ -46,20 +46,64 @@ jobs:
...
+      # E1 static analysis: cppcheck over OUR OWN code (components/ + main/). Fails on
+      # any error/warning/performance/portability finding (--error-exitcode). The C
+      # `error` checks are always on. False positives cppcheck's parser cannot model
+      # (e.g. GNU asm() labels on extern binary-blob symbols) are suppressed inline
+      # at the site with a justification; the missing-ESP-IDF-headers noise and the
+      # normal-check-level branch-limit note are suppressed globally (informational).
+      - name: Static analysis (cppcheck)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cppcheck
```
justinh-rahb, line 76: [P3] Non-blocking: `ubuntu-latest` plus an unversioned apt install makes the new static-analysis gate change over time. Today's run used Cppcheck 2.13.0; a runner-image update can alter pass/fail behavior without a repository change. Pinning the runner/tool environment (ideally a digest-pinned image, or at least a fixed runner and documented package version) would make this gate repeatable.

justinh-rahb:
Hardware follow-up on the exact reviewed head `3cc336d`:

- Re-ran all six C host suites, the 11-case HIL-runner test, API/dashboard checks, and `git diff --check`: **PASS**.
- Reproduced the CI static-analysis command in Ubuntu 24.04 with Cppcheck 2.13.0 across all 16 first-party source files: **PASS**.
- Built default, release-config, native-USB HIL, and UART HIL profiles. Both HIL images passed `check_hil_compileout.sh`, confirming the Panda GPIO/ADC backends are absent.
- Flashed the UART HIL image to a physical dual-USB-C ESP32-C3 dev board through its CH341 bridge and ran the complete suite: **3 scenarios / 66 steps / 49 correlated commands, all PASS**.

The physical run covered AUTO engage/disconnect fail-off, button debounce and mode toggles, panic-off, remote-lease invalidation, chamber-open fail-closed behavior, PTC over-temperature shutdown, fault recovery, and zero-cross injection. Every relevant snapshot reported `io.target: devboard` and `mains_gpio_compiled_out: true`. Final state was `mode: off`, `heater_output: false`, `fault_latched: false`, and `fan_percent: 0`.

So the functional/safety changes and HIL tooling check out on real dev-board hardware. This does not change the dependency/IDF reproducibility and smaller CI-coverage notes in the review above.

plastikman:
## Self-review + polish (commit \`05caed0\`)

Reviewed the whole PR line-by-line. Verified: both safety refactors (\`pb_heater_eval_trip\`, \`pb_ntc_classify\`) are **behavior-identical** to the original tick/read logic (priority order, \`*_ok\` gating, NAN handling all preserved); the two new host tests are correct and thorough (traced every assertion against the inlines — over-temp priority, armed-vs-unconditional gating, NAN-can't-fabricate-a-trip, OPEN/SHORT/rail boundaries, raw-checks-precede-voltage ordering); the Moonraker fixes are value/behavior-identical; `-Werror` is correctly PRIVATE to `pb_*` only. Builds clean default+release+HIL, cppcheck exit 0, all host tests pass, and bench-validated (sensors read `ok`, no false trip, Moonraker connects).

**Applied two review notes:**
- **CMake `-Werror`** now auto-discovers the `pb_*` components (no hardcoded list → new components covered automatically) and **asserts at least one matched** — if ESP-IDF's `__idf_<name>` target convention ever changes, the configure fails loudly instead of silently applying `-Werror` to nothing. (Prints the count; currently 13 in the default build, `pb_hil` correctly skipped.)
- **`copy_upper`** guards `dst_sz==0` (the `dst_sz-1` unsigned wrap); no caller passes 0, this just makes it safe regardless.

**Deliberately left as-is:** cppcheck is installed via apt and not version-pinned. Pinning to an exact apt version trades "a runner upgrade could surface new findings" for "install fails when the image drops that version" — neither is clearly better, and the inline-suppression approach already absorbs version drift. Flagging for future consideration, not blocking.

PR #30, opened 2026-07-26
feat(heater): element-temperature foldback limiter below the 105 °C cutoff

plastikman:
## What

On a hot/marginal install, the bang-bang heater drives the PTC element full-power until it self-heats to the **105 °C cutoff**, which **latches a fault** and forces a manual clear — repeatedly, often before the chamber ever reaches set-point. (This is the customer "get to ~55, PTC hits 105, shut off, press clear 3–4×, repeat" report.)

This adds a **soft foldback layer strictly below the hard cutoff**: once the element enters `[100 °C, 105 °C)` the SSR duty is linearly reduced toward 0 as it approaches 105, so the element **holds just under the cutoff** and the chamber keeps warming instead of tripping.

```
duty
100% ┤━━━━━━━━━━━━●            (ptc ≤ 100 → full power)
     │            ╲
 50% ┤             ╲           (ptc 102.5 → 50%)
     │              ╲
  0% ┤               ●──       (ptc → 105 → 0%; hard cutoff owns ≥105)
     └──┬────┬────┬──┬──
      100  102 103 105 °C
```

Since the MGR GJ-5-L is a **zero-cross SSR** (on/off, no phase-cut), the duty is applied by **time-proportioning** — a first-order sigma-delta across the 2 Hz control ticks (at 50 % it toggles each ~500 ms tick; at ~30 % it fires ~1 tick in 3).

## Safety — unchanged, defense-in-depth preserved

- The **hard 105 °C PTC cutoff stays the first, unconditional, latching check** in `pb_heater_tick`. The foldback only shapes power *below* it.
- The foldback can **only ever reduce** the base bang-bang demand — it never adds heat and never delays the trip.
- If the element still reaches 105 °C (welded SSR, runaway, sensor fault), it **latches off exactly as before**. This only prevents *nuisance* trips on marginal-but-healthy setups.
- The 85 °C chamber cutoff, sensor-fault trips, and comms deadman are untouched and still evaluated first.

## How

- `pb_heater.h`: `PB_HEATER_PTC_FOLDBACK_START_C` (**100 °C**, the single compile-time tuning knob) + a **pure, host-testable** inline `pb_heater_foldback_duty(base_duty, ptc_c)`. Retuning the threshold is a one-line change; the tick, the duty math, and the host test all derive from it.
- `pb_heater.c`: split the chamber bang-bang into a steady **heat-intent latch** (`s_heat_intent`) vs the **momentary SSR state** (`s_on`); apply the foldback cap; drive the SSR via a **sigma-delta accumulator** (`s_duty_accum`). State resets when disarmed/latched. `pb_heater_heat_mode()` (fan/LED signal) stays steady — only `pb_heater_is_on()` (documented momentary) sees the modulation.

## Bench validation (v0.6.3-rc1…rc4 pre-releases, real hardware)

- **Board:** validated on a **Panda Breath V1.0.1** board (boot, OTA, heat cycle, and the foldback). A **V1.0** board looks functionally identical (same PSU/SSR/NTCs/pinout) and is expected to work, but that's **not yet hardware-confirmed** — awaiting feedback from a V1.0 user.
- **Normal heating works** after the bang-bang refactor — element self-limits ~82 °C with full airflow, SSR steady full-on below target; foldback correctly dormant below the band. ✅
- **Foldback engages on real silicon:** with the blower intake restricted, the element climbed into the band and the **instant it crossed the threshold the SSR started cutting (`out=False`)**, trimming it back. Under the hardest case tested — **100 °C heated bed loading the chamber + 70 °C chamber target** — the foldback held the element steady in-band **at ~73 % SSR duty with no 105 °C trip**. That same condition on the old bang-bang firmware drove the element straight to 105 and a latching fault. ✅
- **Threshold = 100 °C (5 °C band).** Bench testing at 102 °C held fine on the V1.0.1/82 kΩ board, but the V1.0/33 kΩ board was seen running closer to the cutoff, so the start is set to **100 °C** to give every board a 5 °C bleed-off margin — the small loss of top-end heating on the 82 kΩ board is negligible. (Kept a single global value rather than tying it to the Rref variant, so this PR stays independent of the strap-detection PR.)
- **Board coverage:** all of the above is on **V1.0.1**. V1.0 hardware confirmation is pending a user with that board.

## Testing

- Host tests: `pb_heater_foldback_duty` band edges / linear ramp / min-not-overwrite / base-0 passthrough, all **derived from the define** — PASS. All other host suites still pass.
- Full ESP-IDF build green; OTA'd to the bench and exercised live (see above).

## Notes for review

- Foldback start is a fixed 100 °C `#define`. Easy to promote to a persisted runtime setting (or a per-Rref-variant value) later if per-install tuning is wanted.
- The RE finding stands: stock does **not** modulate — it holds the **blower continuously** while heating. This foldback is complementary (nuisance-trip mitigation); ensuring continuous blower airflow at set-point is a separate lever worth its own look.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
+### Added
+- **Element-temperature foldback limiter.** Instead of driving the heater full-power
+  into the 105 °C PTC-element cutoff and hard-faulting, the SSR duty is now linearly
+  folded back once the element climbs into `[100 °C, 105 °C)`, so it *holds* just under
+  the cutoff and the chamber keeps warming instead of tripping. On a hot/marginal
+  install this replaces the "trip → clear → trip again" loop with a stable (if slower)
+  approach to set-point. Because the SSR is a zero-cross type, the duty is applied by
+  time-proportioning (a first-order sigma-delta) across the 2 Hz control ticks. **Safety
+  is unchanged:** the foldback can only ever *reduce* power, and the hard 105 °C cutoff
+  remains the first, unconditional, latching check — if the element still reaches it
+  (welded SSR, runaway, sensor fault) it latches off exactly as before. The pure
+  foldback curve is covered by host tests.
+
 ## [0.6.2] - 2026-07-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ below into the GitHub Release notes.
 ### Added
 - **Element-temperature foldback limiter.** Instead of driving the heater full-power
   into the 105 °C PTC-element cutoff and hard-faulting, the SSR duty is now linearly
-  folded back once the element climbs into `[100 °C, 105 °C)`, so it *holds* just under
+  folded back once the element climbs into `[102 °C, 105 °C)`, so it *holds* just under
   the cutoff and the chamber keeps warming instead of tripping. On a hot/marginal
   install this replaces the "trip → clear → trip again" loop with a stable (if slower)
   approach to set-point. Because the SSR is a zero-cross type, the duty is applied by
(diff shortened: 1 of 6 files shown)


plastikman:
Superseded by the element-foldback **hysteresis** (the proportional ramp here proved too gentle on hot boards — ratcheted to 105 and tripped). Shipped in v0.6.3.

PR #31, opened 2026-07-26
fix(board): robust Rref strap read — tolerate a floating GPIO19

plastikman:
## Symptom

A user flashed the firmware and the DragonBreath chamber reading **instantly dropped from ~27 °C to 12 °C** while the room and every other printer thermistor stayed at 26–27 °C. **Both** the chamber and PTC sensors read the *same* wrong value simultaneously. A couple of **hard resets** restored the correct 27 °C.

## Root cause — a floating Rref strap, not a firmware change

`pb_ntc` and `pb_board` are **byte-identical** across the versions involved (verified: zero commits touch them between v0.6.1 and the flashed build) — no calibration, R25, beta, or conversion change. The tell is that **both channels shift by the same amount**: the only thing they share is the **reference resistor** `Rref`, selected by a strap on **GPIO19** and read once at boot.

That read had **both internal pulls disabled**, so on a board that doesn't firmly drive GPIO19 the pin **floats** and latches an arbitrary level — and it can latch **differently on a cold power-on vs a warm OTA reboot**:

- v0.6.1 installed by **USB flash (cold boot)** → right Rref → 27 °C
- newer build installed by **OTA (warm reboot)** → wrong Rref → 12 °C
- **2× hard reset (cold)** → right Rref → 27 °C again

"Hard reset fixes it" rules out any *stored* value (NVS survives a power cycle); it's re-evaluated every boot. A too-large Rref (82 kΩ used where 33 kΩ is correct) computes the NTC resistance ~2.5× high → reads cold, ~15 °C low. This also mildly **degrades safety** on the affected board: cold-biased readings make the over-temp cutoffs trip *late*.

## Fix — dual-pull float detection

Sample GPIO19 under an internal **pull-up** and an internal **pull-down**:

- **Firmly strapped** pin → driven by the board → reads the **same** both ways → **trust it** (unchanged behavior).
- **Floating** pin → follows the pull → the two reads **disagree** → **don't trust it** → fall back to the **fail-safe 33 kΩ** default.

33 kΩ is deliberate: a *smaller* Rref biases both readings **warm**, so the fixed 105 °C/85 °C cutoffs trip **early, not late** — the safe direction — and it's the value the observed V1.0 board actually needs.

Also exposes the resolved reference at **`GET /api/v2/info` → `rref_kohm`** so a board's Rref (and whether it floated) is visible for diagnostics.

## Validation

- **V1.0.1 bench (firm strap):** dual-pull reads identically → `rref_kohm = 82` (documented value); temperature **unchanged across the flash** (40.9→40.7 °C, smooth) — the fix is a no-op on firmly-strapped boards. ✅
- A **floating V1.0 board** will now resolve deterministically to `rref_kohm = 33` regardless of cold/warm boot.
- Full ESP-IDF build green.

**For the affected user:** after flashing this, check `http://<device>/api/v2/info` — `rref_kohm` should read `33` and the chamber temp should track the room again, consistently across both cold and OTA reboots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
+### Fixed
+- **Rref strap misread on a floating GPIO19 (both NTC readings ~15 °C off).** The
+  thermistor divider's reference resistor (82 kΩ vs 33 kΩ) is selected by a strap on
+  GPIO19, read once at boot — previously with both internal pulls disabled. On a board
+  that doesn't firmly drive the pin it **floats** and can latch the wrong value,
+  **differently on a cold power-on vs a warm OTA reboot** — silently choosing the wrong
+  Rref and shifting **both** the chamber and PTC readings by ~15 °C (observed on a V1.0
+  board: 27 °C after a USB flash, 12 °C after an OTA update; a couple of hard resets read
+  27 °C again). The strap is now sampled under an internal pull-up **and** an internal
+  pull-down: a firmly strapped pin reads identically both ways and is trusted, while a
+  floating pin disagrees and falls back to the **fail-safe 33 kΩ** default (a smaller
+  Rref biases readings *warm*, so the fixed 105 °C/85 °C over-temp cutoffs trip early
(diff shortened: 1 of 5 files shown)


plastikman:

PR #35, opened 2026-07-27
feat(ota): upload % on /fw + auto-return to dashboard after update

plastikman:
Two small OTA-page UX asks.

## Upload % complete
The `/fw` page used a plain `fetch` (no progress) — you just saw "Uploading…" then "Rebooting". It now uses `XMLHttpRequest` so `upload.onprogress` can show a **live percent** while uploading/flashing. Because the device writes each chunk to flash as it's received, the upload % tracks the flash itself. On upload completion it shows "Verifying image…".

## Auto-return to the dashboard
After the image is accepted, `waitBack()` polls `/api/v2/info` every 2 s and, once the rebooted device answers on the new firmware, **redirects to the main screen** (`/`) — no more manual "Back to status". ~2-minute fallback redirect if it can't confirm.

## Error handling
A connection drop **after** the upload finished (`up` flag set) is the expected reboot → treated as success + `waitBack()`. A drop **mid-upload** is reported as a **retryable failure** and the button re-enables — so a genuine network glitch isn't mislabeled "Flashed".

## Validation
- Firmware builds clean.
- The embedded page JS was syntax-checked (esprima parse, clean).
- ⚠️ The visual % bar + redirect are browser behaviors — needs a quick **manual check in a browser** (flash any build via `/fw`, watch the % climb, then confirm it lands back on the dashboard). I can OTA this to the bench once it's idle so it's live to test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
+### Changed
+- **OTA update page shows upload progress and returns to the dashboard automatically.**
+  The firmware-update page (`/fw`) now streams the image via `XMLHttpRequest` and shows a
+  live **percent complete** while uploading/flashing (the device writes bytes as it
+  receives them, so the upload % tracks the flash), then — once the image is accepted —
+  **polls the rebooting device and redirects to the main screen** when it comes back on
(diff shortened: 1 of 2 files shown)


plastikman:

PR #36, opened 2026-07-27
feat(ui): dashboard chart — overlay chamber + PTC with labeled axes

plastikman:
Dashboard temperature chart improvements (requested).

## Both series on one graph
The trend previously plotted only the **chamber** temperature (a faint sparkline behind the big number). It now overlays **both chamber and PTC element** on a shared, auto-scaled graph so you can see the element climbing relative to the chamber set-point — handy for watching the foldback hold the element under the 105 °C cutoff.

## Labeled axes + legend
- **Y axis**: °C max/min of the current auto-scale (left).
- **X axis**: the time span back from "now" (e.g. "−4 min").
- **Legend**: Chamber (blue) / PTC (orange).

The chart moved from a faint absolute backdrop to a dedicated in-flow block below the readouts. A `null` sample breaks that series' line (no draw-through-zero).

## Validation
- Firmware builds (app.html re-gzips into flash).
- Dashboard `<script>` syntax-checked clean (esprima).
- ⚠️ **Layout is browser-only** — I can't render it headlessly. Needs a visual check on the bench once the current print frees it up (~45 min). Layout direction was picked from a mockup ("dedicated labeled chart"); happy to tweak spacing/heights after you see it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
+### Changed
+- **Dashboard temperature chart overlays chamber + PTC element with labeled axes.** The
+  trend plotted only the chamber temperature (a faint sparkline behind the reading). It
+  now draws **both** the chamber and the PTC-element temperature on one shared, auto-scaled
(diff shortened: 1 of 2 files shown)


plastikman:

PR #53, opened 2026-07-30
RFC: no-USB WiFi installer via the stock updater

plastikman:
**Design doc for review** (docs-only) — installing DragonBreath over WiFi using the stock firmware's own web updater, with a device-generated stock backup as the safety net. No USB, no opening the unit.

## Why
USB flashing is the top adoption barrier, and the new Panda has **no external USB**. Every Panda already runs stock firmware with a built-in web "Firmware Update" — push DragonBreath through that.

## RE established (from the 1.0.3 stock image, local only)
- Stock has a **local-`.bin` web OTA** (`esp_ota`), reachable at `PandaBreathe.local`.
- **No secure boot / no flash encryption** (proven — we boot unsigned images).
- **No image-identity gate** (only standard `esp_image` checks).
- Dual-OTA partition table: **app slots 1920 K @ 0x10000 / 0x1f0000** — our app (~1.05 MB) fits easily.

## Plan
A small **stock-layout Installer app** flashed via stock's updater, which:
1. **Backs up the full 4 MB stock flash to a download** (read-only, zero risk) — the safety net.
2. **Installs DragonBreath into the inactive slot** (Option A: no bootloader/partition-table rewrite → no self-overwrite, rock-solid). Because the stock bootloader/table are untouched, **WiFi restore is also a safe app-OTA**, closing the Panda→DragonBreath→Panda round-trip.

Option B (full native convert) is documented but **not** recommended for v1 (self-overwrite/brick risk); USB stays the power-user native path.

## Hard licensing line
**We ship only our own code.** The stock backup is generated by the user's own device from its own flash — never redistributed by us. `pb_full.bin` and RE artifacts are local-only.

## Open questions to resolve before building
Stock-bootloader ↔ our-app boot compatibility (top test item), NVS coexistence, slot occupancy, backup completeness. Testing uses a locally-restored stock unit.

Doc: `plans/webflash-installer.md`

### DIFF
--- a/plans/webflash-installer.md
+++ b/plans/webflash-installer.md
@@ -1,0 +1,169 @@
+# RFC: WiFi installer — no-USB DragonBreath onboarding via the stock updater
+
+Status: **Draft / design — for review.** A way to install DragonBreath over WiFi
+using the stock firmware's own web updater, with a device-generated stock backup as
+the safety net, so a non-technical user never opens the unit or touches USB/esptool.
+
+## Motivation
+
+USB flashing is the single biggest barrier to adoption. It means opening the unit,
+installing esptool, and running `tools/flash.py`. The **newer BIGTREETECH Panda
+revision has no external USB at all**, so the current path requires physically
+opening the case — a non-starter for most users ("our users are smart, but the
+general population is not").
+
+Every Panda already runs stock firmware with a built-in web updater. If we can push
+DragonBreath *through that*, onboarding becomes: open a web page, upload one file,
+done.
+
+## What the reverse-engineering established (from the 1.0.3 stock image)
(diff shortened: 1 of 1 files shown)


plastikman:
Closing as **done** — the no-USB install path this RFC proposed is now the shipped, recommended install method. Every release (v1.0.2 → v1.0.4) installs `dragonbreath-<ver>.bin` as an app through the stock Panda firmware's own web OTA updater, with the stock bootloader/partitions left in place so users can revert to Panda over WiFi. See the install instructions on any recent release. The mechanism REd here is realized in `tools/package_release.sh` + the release notes; no separate installer artifact was needed.

PR #60, opened 2026-07-31
RFC: no-USB Panda -> DragonBreath migration (paxx 1.6.0)

plastikman:
Plan for **converting existing stock-Panda users to DragonBreath over the network (no USB)** when paxx 1.6.0 removes the Panda integration.

## Why it's forced, not optional
1.6.0 deletes the "Panda Auto" path. Any user still on **stock Panda firmware** has a **dead chamber heater** on update unless we convert them — device firmware *and* Klipper config — seamlessly.

## The key finding (the linchpin)
Stock stores WiFi in `app_nvs/wifi_info` (blob) and Moonraker in `moonraker_info` (blob) — **same NVS namespace** DragonBreath uses, but DragonBreath reads separate `ssid`/`password`/`mk_host` keys. *That's why WiFi didn't carry over on the bench.* So seamless carry-over is a **small DragonBreath first-boot shim** (parse the stock blobs → populate our keys), not a paxx-only problem. Verified by parsing a real stock backup's NVS.

## Three parts
- **DragonBreath firmware** — the NVS carry-over shim (linchpin): OTA flip rejoins the same WiFi + Moonraker with zero re-provisioning.
- **paxx orchestrator** — detect/classify → network-flash the pinned image via stock's own OTA endpoint (stock stays in the **inactive slot** as rollback) → verify → swap Klipper config (carrying the user's settings) → restart klippy → verify live. Idempotent, resumable, offline-capable (bundled image).
- **dragonbreath-klipper** — discovery/health helper + generated config.

## Safety (it's a forced flash)
Stock stays bootable via **boot-inactive**; dual-OTA rollback on a bad image; **staged config** (keep stock config until DragonBreath verifies); **fail loud, not silent** (a quietly-dead heater after a forced update is the worst outcome).

## Open questions
Exact stock firmware-update network call (HTTP `/ota` vs `/ws`), `wifi_info`/`moonraker_info` blob layouts, DHCP IP stability across the mDNS rename, settings mapping, consent model (one-tap recommended), version pinning.

### DIFF
--- a/plans/panda-to-dragon-migration.md
+++ b/plans/panda-to-dragon-migration.md
@@ -1,0 +1,151 @@
+# RFC: Seamless no-USB migration — stock Panda Breath → DragonBreath (paxx 1.6.0)
+
+Status: **Draft / design — for review.** paxx 1.6.0 (paxx12) merges DragonBreath and
+**removes the stock-Panda chamber-heater integration entirely**. Any user still running
+**stock Panda firmware** with the old "Panda Auto" config therefore has a **dead chamber
+heater** the moment they update — unless 1.6.0 converts them. This RFC is how we convert
+them **automatically, over the network (no USB), safely, with a guaranteed fall-back to
+stock.**
+
+## Why this is a forced conversion, not an offer
+
+1.5.2-paxx12 shipped **both** paths (Panda Auto + DragonBreath) in the settings
+dropdown. 1.6.0 deletes the Panda path (paxx will not maintain two integrations). So on
+update, an existing stock-Panda user's chamber heater stops being controllable until the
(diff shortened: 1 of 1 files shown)


plastikman:
Closing as **done** — the no-USB Panda→DragonBreath migration this RFC proposed has been implemented and shipped in the paxx firmware (SnapmakerU1-Extended-Firmware PR #626). Users move from the stock Panda chamber-heater to DragonBreath over WiFi via the stock web updater — no USB — with the chamber-heater dropdown migration handled in the paxx overlay. The mechanism designed here is realized there; no further work needed on the DragonBreath side.

Issue #64, opened 2026-08-05
Feature Request: Heating Zones


Egosumumbravir:
First of all, enjoy your holidays!

Secondly, fantastic work!

Thirdly, being able to define thermal zones in the automation would be awesome.

Why?  Because PETG like a warm chamber and having the Dragon bring it up to 40°C for a PETG print makes for stronger parts. There is not much difference between PETG and ABS bed temperatures though.

Even PLA has temperatures under which it likes to play merry hell with warping but that's probably more difficult.

Just a thought.

plastikman:
I’ll assume you are looking for something more than “follow the bed”. With klipper I can set the heater with a macro at the start of the print, or even set the different filaments to call for a specific temperature before the print. 

Are you looking for this setup with a Bambu printer to be more automatic like it works woth klipper?  I don’t have a Bambu printer, so all the validation I can do is with bambuddy. If the printer exposes the filament to the MQTT we can read that and do more with it, perhaps we can also make the gcode parts work too. 

If you can clarify the requirements a bit more I can get to coding.

plastikman:
This is now implemented and shipped in **[v1.0.4](https://github.com/plastikman/DragonBreath/releases/tag/v1.0.4)** — give it a shot! 🐉

**How it works (Bambu source only):** during a Bambu print, DragonBreath reads the active tray's filament type from the LAN report and sets the chamber target from a per-filament map — e.g. PETG → 40 °C — instead of the bed-threshold AUTO seam. It preheats on `PREPARE`, holds through a `PAUSE`, and reverts to idle/AUTO when the print ends. A target of `0` = "no zone / off", and an unrecognised filament also resolves to 0. (Klipper users are unaffected — that path drives the chamber via `M141`/`M191` macros.)

**Built-in filament types** (each editable): PLA/TPU off, PETG 40, ABS/ASA 55, PC 60. You can set these during initial setup on `/setup`.

**Custom profiles:** for anything not in the built-in set (PA, PCTG, …), the dashboard has a **Custom profiles** card — add up to 8, applied live (no reboot) and persisted across reboots. Longest-prefix matching means a custom `PETG-CF` wins over the built-in `PETG`. A Bambu-only **Filament** row in the status card shows the active print's filament while it runs.

**Install (no USB):** on the stock Panda web UI → *Firmware Update*, upload `dragonbreath-v1.0.4.bin` from the release. If you're already on DragonBreath, upload it via `/fw`.

[...]

PR #65, opened 2026-08-06
Bambu filament-based chamber zones (WIP)

plastikman:
Addresses #64 (heating zones). **Bambu-only by design** — Klipper users already drive the chamber via `M141`/`M191` macros ("the way"), and DragonBreath sits *in* the Klipper gcode path so no auto-detection is needed there. Bambu never sends chamber intent to DragonBreath (it isn't in Bambu's toolchain and ignores `M191`), so the only way to give Bambu users filament-aware chamber temps is for DragonBreath to read the loaded filament off MQTT and apply a filament → temperature map.

## Plan (phased)
- [x] **A — Parse active filament** *(this commit)*. `pb_bambu` extracts the active filament from the report's AMS / external-spool data (`tray_now` → `ams.tray[N].tray_type`, else `vt_tray.tray_type`), exposed as `pb_bambu_status_t.filament`, logged on change.
- [ ] **B — Filament → temp map** — user-configurable, NVS-persisted, with sane defaults (e.g. PETG 40, ABS/ASA 45, PLA 0/off).
- [ ] **C — Apply via AUTO** — when a Bambu print is active with a known filament, drive the chamber target from the map; revert on print end.
- [ ] **UI** — edit the zone map on the setup page.

## Validation
- **Parser: host unit test, 7 cases** — AMS tray 0/1/2, external spool, none-loaded, idle (no AMS), vt fallback — **all pass**.
- Live on the bench bound to a **bambuddy** virtual X1C: connects + subscribes. bambuddy archive-mode emits no AMS block, and its broker won't relay client-injected `/report` messages, so **end-to-end filament→zone will be confirmed on the requester's real printer** (and the pure parser logic is covered by the unit test).

## Notes / open questions
- Bambu ignores `M191` and never emits a "chamber setpoint" — deriving it from filament at our layer is the correct approach (bambuddy does the same internally).
- The map is opinion ("PETG likes 40 °C") → user-configurable with defaults, not hardcoded.
- Multi-material: likely follow the active tray (`tray_now`); open to "highest-temp loaded filament" instead.

Draft while B/C/UI land.

### DIFF
--- a/components/pb_bambu/include/pb_bambu_parse.h
+++ b/components/pb_bambu/include/pb_bambu_parse.h
@@ -1,0 +1,92 @@
+#pragma once
+// Pure, host-testable Bambu report parsing (no ESP deps) — mirrors the pb_ntc /
+// pb_heater convention of keeping decision logic inline in a header so it can be
+// unit-tested on the host. Used by pb_bambu.c and tests/pb_bambu_host_test.c.
+#include <string.h>
+#include <strings.h>   // strncasecmp
+#include <stdlib.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdio.h>
+
+// Result of resolving the active filament from a report. The tri-state matters:
+// a delta report that OMITS filament state must NOT clobber the last known value,
+// whereas an explicit "nothing loaded" MUST clear it (else a stale zone applies).
+typedef enum {
+    PB_FILA_ABSENT = 0,   // no filament state in this (delta) report -> keep prior
+    PB_FILA_EMPTY,        // report explicitly has no active/loaded filament -> clear
+    PB_FILA_PRESENT,      // an active filament type was found -> use `out`
+} pb_fila_result_t;
+
+// Copy the string value of a JSON key ("key":"value") into out. `key` includes the
+// quotes. Returns true if a (possibly empty) string value was found.
+static inline bool pb_bambu_find_string(const char *s, const char *key, char *out, size_t outsz)
+{
+    const char *q = strstr(s, key);
+    if (!q) return false;
+    q = strchr(q, ':');
+    if (!q) return false;
+    q++;
+    while (*q == ' ' || *q == '\t' || *q == '\n' || *q == '\r') q++;
+    if (*q != '"') return false;
+    q++;
+    size_t i = 0;
+    while (*q && *q != '"' && i + 1 < outsz) out[i++] = *q++;
+    out[i] = '\0';
+    return true;
+}
+
+// Resolve the ACTIVE filament type from a Bambu report. Bambu lists loaded filaments
+// under print.ams (AMS units, tray[].tray_type) and print.vt_tray (external spool),
+// with print.ams.tray_now selecting the active global tray index (254 = external
+// spool, 255 = none loaded). Targeted scan — no full JSON parse over ~15 KB.
+static inline pb_fila_result_t pb_bambu_active_filament(const char *json, char *out, size_t outsz)
+{
+    out[0] = '\0';
+    const char *tn  = strstr(json, "\"tray_now\"");
+    const char *vt  = strstr(json, "\"vt_tray\"");
+    const char *ams = strstr(json, "\"ams\"");
+    if (!tn && !vt && !ams) return PB_FILA_ABSENT;   // delta with no filament state
+
+    char trayn[8] = {0};
+    long now = (tn && pb_bambu_find_string(json, "\"tray_now\"", trayn, sizeof trayn))
+                   ? strtol(trayn, NULL, 10) : -1;
+    char tt[16];
+
+    // An AMS slot is selected: report the active tray's type. An empty type there
+    // means that slot has nothing loaded -> EMPTY (do NOT fall back to vt_tray).
+    if (now >= 0 && now < 250) {
+        if (ams) {
+            const char *p = ams; long idx = -1;
+            while ((p = strstr(p, "\"tray_type\"")) != NULL) {
+                if (++idx == now) {
+                    if (pb_bambu_find_string(p, "\"tray_type\"", tt, sizeof tt) && tt[0]) {
+                        snprintf(out, outsz, "%s", tt); return PB_FILA_PRESENT;
+                    }
+                    break;
+                }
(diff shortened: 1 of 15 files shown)


plastikman:
**UX decision:** the filament→temp zone map is shown **only when Bambu is the selected control source** — it lives inside the Bambu group (`data-src=1`) on the setup page, so `srcshow()` auto-hides it for Klipper/HA/None. Klipper users drive the chamber via M141/M191, so surfacing zones there would just be clutter.

components/pb_bambu/pb_bambu.c
```diff
@@ -124,8 +179,11 @@ static void parse_report(const char *json)
     }
     if (got_tgt)  s_status.bed_target = bedtgt;
     if (got_cham) s_status.chamber_temp = cham;
+    bool fila_changed = got_fila && strcmp(fila, s_status.filament) != 0;
```
justinh-rahb: `active_filament()` returns `false` both when a delta omits filament data and when the report explicitly indicates no active spool (`tray_now` 255 / an empty tray type). Because this write is conditional on `got_fila`, a previously selected PETG/ABS value remains in `s_status.filament` after unload or print end. Phase C would then apply the stale material zone.

Please distinguish **unchanged** from **known empty**, clear the stored value for the latter, and add a PETG → no-spool transition test.

justinh-rahb (changes requested):
Requesting changes for the stale active-filament state noted inline.

Also, the PR states that seven host parser tests pass, but no Bambu parser test is included in the PR tree, so those scenarios are not regression-protected. Please add the test alongside the tri-state fix.

The required `firmware-build` check is currently failing; please resolve or explain that failure before merge.

plastikman:
Pushed fixes addressing the review (@justinh-rahb) — ready for another look:

1. **Stale active-filament** — `active_filament()` is now **tri-state** (`PB_FILA_ABSENT` / `EMPTY` / `PRESENT`). A delta report that omits the AMS/tray block keeps the last value; an explicit no-spool (`tray_now` 255, or an empty active tray type) **clears** it, so a stale material zone is never applied after unload/print-end.
2. **Regression test** — added `tests/pb_bambu_host_test.c` + `tests/run_bambu_host_test.sh` + a `firmware-build` step. 13 cases: the tri-state (incl. PETG present → no-spool EMPTY) and case-insensitive filament→zone matching. Pure logic extracted to `pb_bambu_parse.h` (mirrors the pb_ntc/pb_heater host-testable-inline convention).
3. **CI** — the previous `firmware-build` was a *cancelled* job (transient); cppcheck + build were clean locally. This push re-runs it.

Also included: **Phase B** — the filament→chamber-target zone map (defaults PLA 0/off, PETG 40, ABS/ASA 45, PC 50, TPU 0; NVS-backed, user-overridable). Phase C (apply via AUTO) + the Bambu-group UI still to come.

components/pb_bambu/include/pb_bambu_parse.h
```diff
@@ -0,0 +1,92 @@
...
+                if (++idx == now) {
+                    if (pb_bambu_find_string(p, "\"tray_type\"", tt, sizeof tt) && tt[0]) {
+                        snprintf(out, outsz, "%s", tt); return PB_FILA_PRESENT;
+                    }
+                    break;
+                }
+                p += 11;
+            }
+        }
+        return PB_FILA_EMPTY;
```
justinh-rahb: A selected slot without a tray payload is still possible in a delta (for example `{"ams":{"tray_now":"0"}}`). This branch falls through to `PB_FILA_EMPTY`, clearing a previously known filament even though the report did not explicitly show an empty active tray.

Return `PB_FILA_ABSENT` when the selected tray entry/type is missing; reserve `EMPTY` for `tray_now == 255` or an active tray whose `tray_type` is explicitly present and empty. Please add the partial-AMS delta case to the host test.

plastikman: Fixed in f17b7ff. A selected AMS slot whose `tray_type` is missing from the report now returns **ABSENT** (keeps the last known filament); **EMPTY** is reserved for `tray_now == 255`, an active tray with an explicitly present-and-empty `tray_type`, or an explicitly-empty external spool. Added host-test cases: `partial-ams delta` → ABSENT, `ext-selected-no-vt` → ABSENT, `ext-spool-empty` → EMPTY (16/16 pass).

plastikman:
Feature is now complete in this PR — pushed Phase C (apply) + the UI:

**C — apply via AUTO** (`49cfa18`): when a Bambu print is active (PREPARE/RUNNING/PAUSE), the chamber follows the active filament's zone target through `pb_policy`'s AUTO seam. Per the agreed decisions: **active tray**, **zone wins** (engages directly, bypassing the bed threshold + overriding the AUTO target), **preheat on PREPARE**, **revert to idle/AUTO** on print end. `pb_policy_set_env()` gained a clamped `src_target_c`; new policy host-test case covers engage/override/clear/clamp.

**UI** (`3e6fdc0`): filament→temp editor inside the Bambu group (`data-src=1`) — shown only when Bambu is the control source, hidden for Klipper/HA/None. Hardware-validated (renders + persists).

Also: `gcode_state` print-state parsing (pure + host-tested), and zone defaults set to real chamber practice (PLA 0, PETG 40, ABS/ASA 55, PC 60, TPU 0).

All host tests (bambu + policy + HIL), cppcheck, and the firmware build are green locally. End-to-end filament→zone still needs a real Bambu print to confirm (bambuddy archive-mode can't emit RUNNING+AMS).

justinh-rahb:
Correction: the parser review findings are addressed and I found no additional blocking functional issue in the completed zone/policy/UI work.

My prior CI statement was too strong. The current PR head changed after I inspected the cancelled workflow; GitHub CLI/API shows no check status attached to the current head rather than an explicit failing required check. I cannot use that stale run as a merge blocker, so please disregard the CI-gate portion of my earlier comment.

The PR description remains worth refreshing to reflect the completed phases, but that is non-blocking.

justinh-rahb (approved):
Approved: the stale-filament and partial-AMS delta findings are addressed with host regression coverage. No further blocking code issues found in the completed implementation.

PR #66, opened 2026-08-06
RFC: add MQTT Klipper integration

justinh-rahb:
## Summary

Accepts the RFC for a portable MQTT-only Klipper integration mode, reconciled with the verified Moonraker/Klipper contract researched in #67 and the dragon-core extraction landed in #68.

- Keeps native `dragonbreath-klipper` as the preferred integration.
- Defines desired state, telemetry, master power, optional macro writeback, sequence coherence, retained-aware arming, heartbeat timeout, and explicit re-arm.
- Records Moonraker's retained split-status JSON, literal macro-object topic names, availability topic, JSON-RPC correlation/deduplication, and in-memory macro-variable behavior.
- Preserves `DC_SRC_NONE = 3` and assigns MQTT-Klipper `DC_SRC_KLIPPER_MQTT = 4` in the required dragon-core follow-up.
- Establishes `db_klipper_mqtt` as DragonBreath product code after #67's post-extraction rebase.
- Requires a central broker, credentials, generated least-privilege ACLs, and real locked-Klipper validation before release.

## M191 decision

Blocking `M191` is impossible without a real Klipper temperature object. MQTT mode therefore supplies a non-blocking `M191` alias with a visible warning. Hard-error behavior was considered and explicitly rejected; generated configuration must not ship or recommend it.

## Validation

- `git diff --check`
- `tests/check_api_v2_contract.sh`
- `tests/run_policy_host_test.sh`
- Documentation-only RFC; no firmware behavior changes.

#67 remains the implementation PR and still requires its post-#68 rebase, dragon-core enum addition, refreshed CI, and real locked-Klipper hardware validation.

PR #67, opened 2026-08-06
Design: MQTT-Klipper mode implementation (grounds RFC #66)

plastikman:
## What this is

A **grounded implementation design** for the MQTT-only Klipper integration mode, building directly on @justinh-rahb's RFC in #66. Doc-only (`plans/mqtt-klipper-implementation-design.md`) — **for review before any firmware is written.** Targets **v1.0.5**.

Before designing, I ran four documentation sweeps against the *official* Moonraker/Klipper docs (MQTT core, `[sensor]`/`[power]`, macros + printer API, and reference integrations like iHeater / Panda Sense). Several RFC assumptions needed correcting — captured in §0 of the doc.

## Verified corrections to the RFC (the ones that change the contract)

- **Availability topic is `{instance}/moonraker/status`** (retained `{"server":online|offline}`, also the LWT) — not `{instance}/mqtt/...`.
- **Split-status payload is `{"eventtime":…,"value":…}` JSON and is *retained*** (not a raw scalar); the combined `.../klipper/status` topic is *non-retained* → we require `publish_split_status: True`.
- **Object names keep their literal space**: `.../gcode_macro DRAGONBREATH/…`.
- **API is correlated by a globally-unique JSON-RPC `id`** on a shared response topic; add **`mqtt_timestamp`** (or QoS 0/2) to avoid double-executing `printer.gcode.script`. Keepalive is a fixed 60 s.
- **Blocking `M191` is impossible by construction** — Klipper expands a macro fully before executing (no await), `TEMPERATURE_WAIT` needs a *hardware* Klipper sensor, and no MQTT-fed Klipper sensor exists. Bonus: `M141`/`M191` aren't native Klipper commands, so we *provide* M141 (no `rename_existing`).
- **`SET_GCODE_VARIABLE` is in-memory only** → device re-publishes desired-state + re-asserts `seq` on reconnect.
- **Moonraker `[sensor]` is invisible to Klipper macros** → dashboard temp and macro-visible temp are two separate paths.

## Design highlights

- **`pb_klipper_mqtt` is a *controller*** (mirrors `pb_ha`: holds a `pb_policy` lease + heartbeats it), reusing the existing lease/policy and `pb_heater` comms-watchdog seams — **no `pb_policy` signature change**, hardware fail-off unchanged.
- **Exactly-one-source stays structural**: a 5th value in the single `/setup` selector, labeled "Klipper (MQTT)" vs "Klipper (Moonraker)", mutually exclusive with the native extra.
- **Retained-aware arming** (the safety core): because split-status is retained, a reconnecting device is handed stale `armed=1`/`target`/`seq` instantly — so heat arms **only** on a live arm-edge *after* connect **plus** a fresh heartbeat (3×5 s = 15 s timeout → force off + latch `comms_lost`). A retained `armed=1` never re-arms.
- **`/setup` = single source of truth → generated config**: collect broker + `instance_name`, derive all topics, emit exact `moonraker.conf` / `printer.cfg` / `mosquitto` ACL; validator refuses to enable without broker + creds.

## Locked decisions (§10)

- **M191** → non-blocking alias + warning (strict hard-error variant shipped commented-out).
- **Writeback** → off by default (opt-in).
- **Topic base** → fixed `dragonbreath`, advanced-editable.
- **`emergency_stop`** escalation → disabled by default.

## Feedback wanted

1. The **retained-aware arming state machine** (§5) — is the "live arm-edge + fresh heartbeat, never trust retained" model airtight?
2. The **controller-vs-bed-follow** choice (§2) and the lease/heartbeat mapping onto `pb_policy`/`pb_heater`.
3. Anything in the **OpenVent shared-core boundary** you'd want handled differently, given 1.1.0 will target the shared base.
4. The **M191 non-blocking default** (§7) — agree it degrades more gracefully than a hard error?

Closes nothing yet — RFC #66 stays open as the spec; this is the implementation plan derived from it.

### DIFF
--- a/.github/workflows/firmware-build.yml
+++ b/.github/workflows/firmware-build.yml
@@ -61,6 +61,9 @@ jobs:
       - name: Test heater safety-trip decision (over-temp / fail-closed / watchdog)
         run: sh tests/run_heater_safety_host_test.sh
 
+      - name: Test MQTT-Klipper arming state machine (retained-aware arm/heartbeat)
+        run: sh tests/run_db_klipper_mqtt_host_test.sh
+
       - name: Test HIL scenario runner
         run: python3 -m unittest tests/test_hil_runner.py
 
--- a/components/db_klipper_mqtt/CMakeLists.txt
+++ b/components/db_klipper_mqtt/CMakeLists.txt
@@ -1,0 +1,9 @@
+idf_component_register(
+    SRCS "db_klipper_mqtt.c"
+    INCLUDE_DIRS "include"
+    # esp-mqtt to the printer's Moonraker broker; esp_timer for pacing; pb_policy for
+    # the lease/command/snapshot API; esp-tls for optional mqtts. Telemetry/RPC JSON is
+    # built with snprintf (no cJSON). The pure arming core (db_klipper_mqtt_arm.h) is
+    # header-only and host-tested separately.
+    REQUIRES nvs_flash mqtt esp_timer pb_policy esp-tls
+)
--- a/components/db_klipper_mqtt/db_klipper_mqtt.c
+++ b/components/db_klipper_mqtt/db_klipper_mqtt.c
@@ -1,0 +1,453 @@
+// Klipper-over-MQTT control source. Subscribes to the Moonraker MQTT split-status
+// for the DRAGONBREATH (desired-state) and DB_LINK (heartbeat) macros, runs the
+// retained-aware arming machine (db_klipper_mqtt_arm.h), and holds a pb_policy lease
+// while armed — exactly like pb_ha, but commanded by Klipper macros over the printer's
+// own Moonraker broker. Publishes device telemetry for a Moonraker [sensor] and an
+// on/off state for a Moonraker [power] device. See
+// plans/mqtt-klipper-implementation-design.md.
+//
+// Threading: the esp-mqtt event handler (mqtt task) updates arm/connection state under
+// s_lock; db_klipper_mqtt_tick() (control task) evaluates the machine, calls pb_policy,
+// and publishes. esp_mqtt_client_publish() is thread-safe; shared state is guarded.
+#include "db_klipper_mqtt.h"
+#include "db_klipper_mqtt_arm.h"
+#include "pb_policy.h"
+
+#include "esp_log.h"
+#include "esp_timer.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/semphr.h"
+#include "mqtt_client.h"
+#include "nvs.h"
+
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static const char *TAG = "db_km";
+
+#define NVS_NS    "app_nvs"
+#define KEY_HOST  "km_host"
+#define KEY_PORT  "km_port"
+#define KEY_USER  "km_user"
+#define KEY_PASS  "km_pass"
+#define KEY_INST  "km_inst"
+#define KEY_TOPIC "km_topic"
+#define KEY_TLS   "km_tls"
+#define KEY_WB    "km_wb"
+
+#define DEFAULT_PORT      1883
+#define DEFAULT_TLS_PORT  8883
+#define DEFAULT_PREFIX    "dragonbreath"
+#define STATE_PERIOD_US   (2 * 1000 * 1000)    // telemetry + power state ~2 s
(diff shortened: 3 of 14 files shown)


plastikman:
@justinh-rahb — couldn't add you as a formal reviewer (permissions), so flagging here: this is the implementation design derived from your RFC #66, grounded in a documentation pass that corrected a few wire-contract assumptions (see §0). Would value your eyes on the retained-aware arming state machine (§5), the controller/lease mapping (§2), and anything on the OpenVent shared-core boundary ahead of the 1.1.0 shared-base work. No firmware written yet — this is the review gate before Phase 2.

plastikman:
Implementation is now pushed (no longer doc-only). @justinh-rahb ready for your review.

**Phases 1–3 built** (all grounded in the doc-sweep corrections in the design):
- `pb_klipper_mqtt` controller component (mirrors pb_ha's lease/heartbeat), NVS-safe `PB_SRC_KLIPPER_MQTT=4` enum, app_main dispatch.
- Retained-aware arming core `pb_klipper_mqtt_arm.h` — pure/host-tested (26 cases in CI): never arms from retained state, arms only on a fresh `seq` + live heartbeat, 3×5s heartbeat timeout → force off, explicit re-arm after recovery, quick to disarm.
- `/setup` group + validator (refuses to enable without broker+creds+instance), `GET /km-config` generates moonraker.conf + printer.cfg + Mosquitto ACL from the saved settings (M141 shim + non-blocking M191 alias, strict variant commented), dashboard label.

Most valuable review targets: the arming state machine (`pb_klipper_mqtt_arm.h` + its host test), and the controller/lease mapping onto `pb_policy`/`pb_heater`.

**Phase 4** (validation on a real locked Klipper install) is still pending and gates the v1.0.5 tag — the bench has no locked-Klipper broker.

plastikman:
Heads-up on sequencing: this collides with the dragon-core extraction in #68 (both touch the control-source enum + `app_main`/`pb_portal`/`pb_httpd`/`pb_ha`). Plan is to **land #68 first, then rebase this PR onto `dc_*`**. The arming logic (`pb_klipper_mqtt_arm.h` + its 26-case host test) and the lease/policy mapping are the substance and won't change in the rebase, so review of those is still valuable now. CI is green (build incl. the new arming host test).

After #68 merges, the rebase will: (1) retarget `pb_klipper_mqtt` onto `dc_source`/`dc_policy` (staying app-side, like `pb_ha`), and (2) require a small dragon-core change adding `DC_SRC_KLIPPER_MQTT` + a `DC_SRC_MAX` clamp (see note on #68).

justinh-rahb (changes requested):
The rebase/core boundary looks right and CI is fully green, including the ESP-IDF and safe HIL-profile builds. One blocking concurrency issue before merge:

`db_klipper_mqtt.c` says MQTT events update shared state under `s_lock` and the control task evaluates it, but that contract is not currently followed. `handle_data()` writes `s_mrk_online`, `s_power_on`, and `s_last_state_us` without the mutex; `publish_power_state()` writes `s_last_power_pub` from the MQTT callback while `db_klipper_mqtt_tick()` reads/writes those same objects. These are cross-FreeRTOS-task C data races, and cppcheck does not model them.

Please either guard every access consistently, or preferably make the MQTT callback only update locked state/queue flags and let `db_klipper_mqtt_tick()` own pacing, publication, and policy calls. Be careful to release `s_lock` before MQTT publish or `pb_policy_*` calls. A focused host test is not required for the mutex mechanics, but the existing arming test should remain intact.

Non-blocking cleanup while touching it: the PR title/body still describe a doc-only design even though this now ships the implementation.

plastikman:
Thanks for the precise review, @justinh-rahb — the concurrency finding was correct. Full disclosure: this merged on green CI just before your CHANGES_REQUESTED landed, so I've fixed it forward directly on main in **699e183** rather than reopening.

Fix follows your preferred shape: the MQTT event handler now **only** mutates `s_lock`-guarded state and sets a `s_pub_pending` request flag; `db_klipper_mqtt_tick()` owns all pacing, telemetry/power publishing, and `pb_policy_*` calls. `s_power_on`/`s_mrk_online` are read+written under `s_lock` everywhere; the `s_last_*` pacing fields are control-task-only; and `s_lock` is always released before any esp_mqtt publish or `pb_policy_*` call. Connect-time power/telemetry publish is deferred to tick via the flag (availability publish + subscribes stay on the MQTT task — thread-safe, no shared C state). Arming host test untouched and still green; build + cppcheck clean. (Title/body were already retitled off the doc-only description before merge.)

PR #70, opened 2026-08-07
Read-only HA telemetry alongside a Bambu/Klipper source

plastikman:
## Summary

Enables **read-only Home Assistant telemetry alongside** a Bambu/Klipper control source — HA is no longer mutually exclusive with the other sources when used purely for monitoring.

- New `pb_ha_start_readonly()`: publishes MQTT-Discovery **sensors** (chamber/element temp + target + mode) and the retained `state` topic, but **never subscribes command topics and never takes a `pb_policy` lease** — it cannot control, only report.
- `app_main` auto-starts it when an HA broker is configured (NVS `ha_host`) but HA is **not** the selected control source. When HA *is* the source, it starts in full-control mode as before.
- The stale controllable **climate** entity is cleared (empty retained config) in read-only mode, so HA shows no dead buttons — just sensors.
- `pb_ha_tick()` is now pumped whenever the client is up (full or read-only).

Enablement + entity style were the owner's call: **auto-when-configured** and **sensors-only**.

## Why

Fixes @Egosumumbravir's report on #64 — "HA reporting dies after connecting to a Bambu machine." That was the single-source design (selecting Bambu stopped the HA client entirely); now HA keeps reporting read-only while Bambu/Klipper drives control.

## Also (opportunistic, same file)

Fixes a **pre-existing cross-task data race** in `pb_ha` — the same class you flagged in `db_klipper_mqtt` #67: the MQTT connect handler wrote the control-task pacing clock `s_last_state_us` directly. Replaced with a lock-guarded `s_pub_pending` request flag that `tick()` reads+clears. This matters more now that `pb_ha` can run concurrently with another MQTT client.

## Scope / validation

- **App-side only — no dragon-core changes.**
- Clean ESP-IDF v5.3 build, cppcheck clean, policy + arming host tests green.
- Not yet validated on hardware against a live HA broker + Bambu concurrently (would appreciate a sanity check there before/with merge).

### DIFF
--- a/components/pb_ha/include/pb_ha.h
+++ b/components/pb_ha/include/pb_ha.h
@@ -33,6 +33,12 @@ typedef struct {
 
 esp_err_t pb_ha_start(void);
 
+// Start HA in READ-ONLY mode: publishes MQTT-Discovery sensors + a retained state
+// topic (chamber/element temp, target, mode) but never subscribes command topics and
+// never takes a pb_policy lease. Used when HA runs ALONGSIDE another control source
+// (Bambu/Klipper) purely as a monitor. Mutually exclusive with pb_ha_start().
+esp_err_t pb_ha_start_readonly(void);
+
 // Periodic pump: publishes retained state (~2 s) and heartbeats the heat lease
 // (~5 s, well under the comms-watchdog minimum). Call from the control loop when
 // HA is the active source. No-op if the client isn't running.
--- a/components/pb_ha/pb_ha.c
+++ b/components/pb_ha/pb_ha.c
@@ -44,14 +44,23 @@ static pb_ha_config_t           s_cfg    = {0};
 static pb_ha_status_t           s_status = { .state = PB_HA_DISABLED };
 static esp_mqtt_client_handle_t s_client = NULL;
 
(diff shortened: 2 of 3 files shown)


plastikman:
@justinh-rahb small app-side feature (no core changes) — read-only HA telemetry that runs alongside Bambu/Klipper, fixing the #64 reporter's 'HA dies under Bambu'. While in `pb_ha` I also fixed the same connect-handler pacing-clock data race you caught in #67 (now a lock-guarded `s_pub_pending` flag). Would value a concurrency sanity check on `pb_ha` running concurrently with another MQTT client.

justinh-rahb:
Coordination note after comparing this directly with #71: this should land first. #71 will then be rebased onto the resulting main branch. There is one expected manual conflict in components/pb_ha/pb_ha.c because #71 migrates the HA client to dragon-core dc_mqtt. During that resolution we will preserve this PR’s read-only HA behavior and s_pub_pending concurrency fix, while translating the new raw ESP-MQTT publish/subscribe calls to dc_mqtt. The header and app_main changes otherwise merge cleanly. No architectural blocker; concurrent HA plus Bambu/Klipper hardware validation should still be done after integration.

PR #71, opened 2026-08-08
Share and harden MQTT clients

justinh-rahb:
## What changed

- Migrate `pb_ha` and `db_klipper_mqtt` from duplicated ESP-MQTT lifecycle code to `dc_mqtt` from justinh-rahb/dragon-core#4.
- Preserve product-local NVS keys, topics, telemetry, policy leases, retained-aware arming, and heater safety behavior.
- Reject fragmented command payloads rather than parsing partial control messages.
- Publish the retained Klipper-MQTT power state on every broker connection, including the default ON state.
- Generate separate least-privilege Mosquitto users for DragonBreath and Moonraker, with the publish/subscribe directions each client actually needs.
- Stop recommending `mosquitto_passwd -c` on an existing password file.
- Correct the implementation design to match shipped behavior: fresh sequence plus heartbeat arming, reserved fan/purge fields, and temperature-only QoS 0 writeback.
- Extend the static configuration contract checks for the broker-account split and safe password-file instructions.

## Why

PR #67 shipped two kinds of follow-up debt: common MQTT transport remained duplicated in the product repository, and the generated broker configuration could not work with its own directional ACL because Moonraker and DragonBreath were assigned the same user. It also skipped the initial retained power-state publication and documented several behaviors that were not implemented.

The new boundary keeps transport lifecycle in dragon-core and all product policy in DragonBreath.

## Dependency and landing order

dragon-core#4 is merged and tagged `v0.4.0`. This PR now pins `dc_mqtt` to permanent merge commit `f283708268298af1e0e1197b923283c50079f539`.

## Validation

- Full ESP-IDF 5.3 ESP32-C3 firmware build with `-Wall -Wextra -Werror`: pass
- Safe dev-board HIL-profile build: pass
- HIL compile-out audit: pass
- All host state-machine and safety tests: pass
- MQTT-Klipper retained-aware arming tests: pass
- API v2 and dashboard contracts: pass
- HIL runner unit tests: 11 pass
- `git diff --check`: pass

Local `cppcheck` was unavailable; GitHub Actions runs the repository's required analyzer.

### DIFF
--- a/components/db_klipper_mqtt/CMakeLists.txt
+++ b/components/db_klipper_mqtt/CMakeLists.txt
@@ -1,9 +1,7 @@
 idf_component_register(
     SRCS "db_klipper_mqtt.c"
     INCLUDE_DIRS "include"
-    # esp-mqtt to the printer's Moonraker broker; esp_timer for pacing; pb_policy for
-    # the lease/command/snapshot API; esp-tls for optional mqtts. Telemetry/RPC JSON is
-    # built with snprintf (no cJSON). The pure arming core (db_klipper_mqtt_arm.h) is
-    # header-only and host-tested separately.
-    REQUIRES nvs_flash mqtt esp_timer pb_policy esp-tls
+    # dc_mqtt owns the shared ESP-MQTT lifecycle; this product component owns the
+    # Moonraker topic contract, retained-aware arming, and policy mapping.
+    REQUIRES nvs_flash dc_mqtt esp_timer pb_policy
 )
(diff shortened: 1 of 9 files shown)


justinh-rahb:
Coordination note: this PR is part of an active predecessor chain. As the preceding PRs merge, we’ll rebase this branch onto the resulting main branch and refresh any pinned commit references before final review/merge. Please treat the current base and dependency SHAs as provisional.

plastikman:
@justinh-rahb #70 (read-only HA alongside a source) is merged to main (`c75bc03`) — so `pb_ha` now has the read-only mode **and** the connect-handler pacing race fixed (lock-guarded `s_pub_pending`, same pattern as the db_klipper_mqtt fix). You can rebase this MQTT-hardening PR onto current main and fold both `pb_ha` and `db_klipper_mqtt` into the shared/hardened client. Flag anything you want changed in either.

PR #69, opened 2026-08-07
Consume shared capability-aware SPA from dragon-core

justinh-rahb:
## Summary

- consume the Dragon-family SPA from dragon-core `dc_ui` and remove DragonBreath's duplicate `app.html`
- pin all seven consumed core components to the permanent dragon-core `v0.4.0` merge revision, `f283708268298af1e0e1197b923283c50079f539`
- add an additive `/api/v2/info.ui` descriptor for schema, product identity, and display name
- keep HTTP/SSE, setup, OTA/recovery, authentication, favicon, and all DragonBreath policy handlers product-local
- update UI/API ownership checks, provenance, release metadata, and RFC progress documentation

## Compatibility

Existing API v2 routes and fields, NVS data, OTA behavior, and recovery behavior are unchanged. Responses without the UI schema retain the static DragonBreath identity; an unknown future schema leaves the static identity and complete UI untouched.

DragonBreath already emits `power_on`, `auto`, and `drying` from `/api/v2/info.capabilities`; the product contract test now asserts those exact values so capability gating cannot silently become inactive.

## Validation

- all DragonBreath C host tests
- API v2 static contract and shared-SPA contract checks
- clean ESP-IDF v5.3 ESP32-C3 firmware build from a dependency-lock-free temporary tree
- verified all seven core components resolve from the pinned commit
- verified the embedded gzip decompresses byte-for-byte to the pinned SPA source
- both safe dev-board HIL profile builds and compile-out checks

## Dependency / landing status

Coordinated dragon-core PRs #2, #3, and #4 landed first. All seven component manifests and release provenance now pin the permanent `v0.4.0` merge commit; #69 contains both DragonBreath and DragonVent SPA surfaces and no longer references a PR-branch-only SHA.

PR #9, opened 2026-07-23
refactor: make policy the authoritative control state machine

justinh-rahb:
## What

Establishes `pb_policy` as DragonBreath's sole owner of modes, requested/effective targets, controller leases, deadlines, and the canonical control snapshot.

This is the first half of the API v2 series. A stacked follow-up will delete the alpha `/status`, `/target`, `/heartbeat`, and `/reset` contract and expose this state through the versioned JSON API and event stream.

## Why

The existing HTTP handlers write `pb_heater` directly while `pb_policy` only provides fan glue. That leaves no single authoritative answer when Klipper, the dashboard, Moonraker, buttons, safety logic, and watchdogs disagree.

## Changes

- adds OFF, POWER_ON, AUTO, and DRYING policy modes
- adds monotonic control-state revisions and source attribution
- adds device-issued 128-bit remote POWER_ON leases
- rejects stale revisions and stale/superseded heartbeats
- makes explicit OFF unconditional and lease-invalidating
- gates AUTO on a live Moonraker subscription with bed hysteresis
- bounds DRYING to 12 hours and local POWER_ON to 12 hours
- keeps boot state OFF and all leases/deadlines RAM-only
- synchronizes heater faults/watchdog expiry back into the canonical snapshot
- routes the alpha HTTP handlers through policy as a temporary adapter
- adds host tests around the actual `pb_policy.c`
- documents the stacked state-machine/API-v2 delivery split

Persistent fault storage and the broader thermal-purge overhaul remain separate safety work; they are not hidden in this transport-independent refactor.

## Safety detail

Policy computation and actuator application share the same mutex. An OFF command therefore cannot be overtaken by a control tick that computed an older positive target.

## Validation

- `sh tests/run_policy_host_test.sh`
- ESP-IDF v5.3 ESP32-C3 build in the official `espressif/idf:v5.3` container
- firmware size: `0xf7c40` bytes; 47% of the smallest app partition remains free
- `git diff --check`

Hardware validation is still required before merge.

PR #73, opened 2026-08-09
Move provisioning and recovery into dragon-core

justinh-rahb:
## Summary

Completes the portal half of the core extraction after #69:

- makes dragon-core `dc_portal` the owner of the HTTP server, shared SPA, captive DNS, Wi-Fi/AP provisioning, logs, OTA and reset transport
- replaces the legacy `pb_portal` pages and DNS implementation with a small `db_portal` product adapter
- keeps API v2 product-local by registering `pb_httpd` routes on the core-owned server
- supplies the controller schema, auth, heater-off maintenance guard, OTA identity validation, and `app_nvs` reset callback
- preserves `/setup`, `/fw`, `/console`, `/km-config`, `/update`, API v2 routes and existing NVS namespaces/keys
- validates product saves atomically, avoids unchanged NVS writes, and retains stored credentials when blank secrets or source unbinding are submitted
- routes normal and HIL builds through the guarded ESP-IDF preflight, including RISC-V compiler discovery and tag-aware stale Component Manager cache detection
- updates README, RFC, changelog, vendoring and release provenance

## Dependency / landing order

The predecessor, justinh-rahb/dragon-core#5, is merged and tagged `v0.5.0`. All eight dragon-core components are pinned to that release.

## Validation

Exact head `23c8b4e`:

- all policy, buttons, heater, NTC, Klipper-MQTT and portal configuration host suites: PASS
- HIL runner unit tests: 13/13 PASS
- API v2/shared portal ownership contract: PASS
- full ESP-IDF v5.3 ESP32-C3 compile + link: PASS
- tag-aware component-lock validation: PASS; `v0.5.0` resolved consistently for all core components
- safe native-USB ESP32-C3 devboard HIL: 3 scenarios, 66 steps, 0 failures
- HIL compile-out audit: PASS; Panda GPIO/ADC backends absent
- normal firmware size: `0x112590`, 43% of each OTA app slot free

Earlier portal/log validation and real-Panda HIL passed on the predecessor head. The production controller is currently cooling after a run and was intentionally left untouched during this update. Keep this PR draft until the current head receives the focused real-controller portal/OTA regression pass.


justinh-rahb:
Coordination note: this branch is intentionally pinned to dragon-core PR #5 and will be rebased/repinned after that predecessor merges. Local ESP32-C3 compile/link and all host/static contract tests pass. Before marking ready, please give the docs/build/release paths an independent pass for stale portal ownership or old build instructions; then we still need the full local devboard plus real DragonBreath HIL, AP/STA provisioning, OTA/update/revert, auth, reset-guard and recovery-route battery.

justinh-rahb:
CI follow-up: the initial run found three adapter-only cppcheck findings (two explicit port casts and the linker-blob pointer-bound idiom). Fixed in `3d8a124`. Replacement run 31286674729 is fully green: host suites, cppcheck, normal ESP-IDF build/link, API/shared-portal contract, both safe devboard HIL profiles, compile-out check, and artifact upload all passed.

plastikman (commented):
Reviewed from "did the DragonBreath-specific `/setup` behavior survive the portal extraction" (this deletes `pb_portal.c`, where several recently-shipped features lived). The schema-driven `db_portal` adapter is a clean shape — but one regression needs restoring before merge.

**Preserved ✅**
- `register_product_routes()` → `pb_httpd_register()`, so all API v2 routes register on the core server — including **`/api/v2/zones`** (GET/POST → `dc_bambu_zone_get_all/set/remove`). So Bambu zones + custom profiles stay fully editable via the dashboard/API.
- `apply_product()` persists every source's config, including the complete **Klipper-MQTT `km_*`** set (host/port/user/pass/inst/topic/tls/writeback) and the `ctl_src` selector (range-checked against `DC_SRC_MAX`).
- Auth, heater-off maintenance guard, OTA identity validation, `app_nvs` factory reset, favicon — all wired.

**Blocking regression 🔴 — the `/km-config` generator is gone**
`db_portal.c` has no `/km-config` route or generator, and `register_product_routes()` only adds `pb_httpd_register` + favicon. That endpoint — the Klipper-MQTT setup helper that generates `moonraker.conf` + `printer.cfg` + the Mosquitto ACL from the saved settings (added in #67) — **shipped in `v1.1.0-rc1`**, so this PR would silently remove a just-released feature. It's product-local and self-contained; please re-register it in `register_product_routes()` (it doesn't need core changes).

**Minor (feature still reachable, low priority)**
- The **`/setup` Bambu zones editor** (the per-filament initial-config inputs) isn't carried over. The dashboard + `/api/v2/zones` still cover it, so not lost — just note if dropping the `/setup` editor is intentional. (An AUTO-mode caption note I'd just added to that page on `main`, `d7cb3b7`, becomes moot — fine.)
- **Explicit Unbind** (old `/unbind`: clear the bound source's creds *and* set source to None) is reduced to selecting "None" via the schema; the stored creds would linger. Confirm that's acceptable or add a clear-on-unbind path.

Draft/landing order looks right (gated on dragon-core#5, repin on merge). CI green. Just the `/km-config` restore is the must-fix from our side.

plastikman (commented):
Additional findings from a code-review pass on the `db_portal` adapter (separate from the `/km-config` regression above), most-severe first:

**1. `apply_product()` is non-atomic — a rejected save can still switch the control source.** (`db_portal.c:~177`)
POST with `ctl_src="1"` (Bambu) + `ha_port="99999"`: `ctl_src` validates and `dc_source_set()` **persists the new source** (and moonraker/bambu sections commit) *before* `ha_port` fails `port_json` → `goto invalid` returns "Invalid source or port value." The user is told the save failed, but the active source (and other configs) are already written to NVS — a restart brings up the wrong source. Suggest validating all fields first, then committing, or applying `ctl_src` last.

**2. Backend-failure paths don't set the `message` buffer → blank/garbage error to the user.** (`db_portal.c:~188`)
If any `*_set_config()` / `dc_source_set()` returns non-OK, `apply_product` returns immediately without ever `snprintf`-ing `message`, so `dc_portal` surfaces an empty/uninitialized error for a real NVS write failure (unlike the `invalid:` path, which does set one).

**3. Redundant NVS writes on every save → needless flash wear.** (`db_portal.c:~186`)
The SPA echoes current values back on submit, and `port_json` sets `present=true` whenever the field is *present* (not whenever it *changed*), so `changed=true` fires and moonraker/ha/km configs are rewritten to NVS on every save even when nothing changed.

**4. (Plausible) `/favicon.ico` may be shadowed by the SPA catch-all.** (`db_portal.c:~279`)
The old `pb_portal` deliberately registered `/favicon.ico` **before** the `/*` catch-all ("before the catch-all so /favicon.ico != SPA"). Now the catch-all is owned by `dc_portal` and favicon is registered via `register_product_routes`; with `httpd_uri_match_wildcard`, if core registers `/*` before invoking the callback, `GET /favicon.ico` gets the SPA HTML. Can't confirm here — depends on registration order inside the pinned `dc_portal`.

(#1 is the one I'd fix before merge; #2/#3 are easy hardening; #4 depends on dc_portal's ordering.)

justinh-rahb:
Hardware qualification update for exact head `e63db5a` (dragon-core pinned to `203164a`):

### Local ESP32-C3 devboard

- Fresh native-USB HIL build and flash at exact head.
- Full suite passed: AUTO/connectivity fail-closed behavior, front-panel button/lease behavior, chamber sensor fault, fixed PTC over-temperature cutoff, fault clear, and zero-cross injection.
- HIL profile confirmed mains GPIO compiled out; final state was safe.
- Report: `hil-results/pr73-devboard-final/report.json` (local test artifact).

### Real Panda Breath controller

- Updated only through the firmware OTA endpoint; OEM bootloader and partition table were not touched.
- Panda serial smoke passed through `pi@klipper:/dev/ttyUSB0`: real target identified, physical zero-cross activity observed, sensors healthy, Moonraker connected, and unconditional OFF accepted.
- Shared SPA aliases `/`, `/setup`, `/fw`, `/console`, and `/diag` all returned the same SPA successfully.
- Provisioning schema returned the DragonBreath product adapter and existing Klipper/Moonraker values; a same-value product save succeeded and preserved the secret fields.
- Authorization behaved correctly: unauthenticated logs returned 403; reset without the exact confirmation returned 400.
- The first authenticated log test found a real ESP32-C3 HTTP-task stack overflow in core. That was fixed in dragon-core `203164a`; after repinning, five consecutive authenticated snapshots returned 200 with an unchanged boot ID.
- With supervision, requested 40 C: heater output and fan asserted, sensors remained healthy, and no fault/inhibit appeared.
- While actively heating, OTA and confirmed factory reset were both rejected with HTTP 409 and the expected safety messages; authenticated logs still returned 200.
- Unconditional OFF returned heater demand/output false and fan off. The remembered manual target was restored to its pre-test 60 C value.
- Finally OTA-restored the clean production profile (`e63db5a`, HIL disabled, Power LED enabled). It booted OFF, fault-free, retained the 82 kOhm board identity and original Klipper/Moonraker/network settings, and served authenticated logs without rebooting.

Not exercised on the real unit: a factory reset / loss-of-credentials AP captive-portal flow, because that would deliberately erase or replace the supervised controller's working network configuration. Station-mode provisioning, schema, auth, recovery aliases, OTA, logs, and operation guards were all exercised.

DragonBreath CI is still running as of this comment; the paired dragon-core CI is green.

justinh-rahb:
Follow-up: final `firmware-build` completed green at exact head `e63db5a`; all host tests, static analysis, ESP-IDF build, shared UI/API checks, and safe devboard HIL-profile build passed.

plastikman (commented):
Re-review of `e63db5a`: this update is a **pin bump only** (`dc_portal` → `203164a`, "hardware fix"); `db_portal.c` is unchanged, so the earlier adapter findings still stand (`/km-config` generator still dropped; `apply_product` still non-atomic — `dc_source_set` at :177 precedes `ha_port`/`km_port` validation at :198/:207; blank error-message paths; redundant NVS writes).

**🔴 New/critical — the repinned `dc_portal` handler-budget fix does not cover DragonBreath, and it's a startup crash (not a CI-visible one):**

`dc_portal` now sets `max_uri_handlers = 16 + config->product_route_count`. But `db_portal` registers its routes via the **`register_product_routes` callback** and leaves `product_route_count = 0`. Concretely at this head:
- `pb_httpd_register()` registers **18** URI handlers,
- `db_portal` adds `/favicon.ico` (1),
- `dc_portal` builtins + `/*` catch-all (~7),

…all against a cap of **16**. The 17th `httpd_register_uri_handler` returns `ESP_ERR_HTTPD_HANDLERS_FULL` → `register_product_routes` fails → `dc_portal_start` `goto fail` → **no HTTP server** (dashboard, `/setup`, API v2, OTA all unreachable). It builds fine, so CI stays green; it fails at runtime — the devboard HIL / real-hardware pass will catch it.

Fix options: (a) `db_portal` declares its routes via the counted `product_routes[]` array so the budget includes them; (b) `db_portal` sets `httpd.max_uri_handlers` explicitly (≥ ~28); or (c) `dc_portal` sizes the table to also account for callback-registered routes. (a) or (b) is the smallest change.

Net for #73 to land: restore `/km-config`, fix the handler budget, and (nice-to-have) the atomic-apply + error-message items.

justinh-rahb:
Addressed the adapter reviews in rebased head `3f4010f`:

- Restored the shipped product-local `GET /km-config` generator, including `moonraker.conf`, `printer.cfg`, the non-blocking M191 behavior, and least-privilege Mosquitto ACL output. The Klipper-MQTT setup section points users to it, and README/FEATURES now document it.
- Reworked product saves into parse -> validate/stage -> persist. Every field is validated before the first setter can touch runtime/NVS, and `dc_source_set()` is deliberately last, so the reported-invalid `ctl_src=1` + `ha_port=99999` case cannot switch sources or partially save earlier sections.
- Added subsystem-specific backend failure messages.
- Added field-level change detection: a normal SPA echo of unchanged values performs no config/source writes. Empty secret fields still retain the saved secret.
- Per Justin's decision, selecting **None** only unbinds the active source enum; all source credentials are retained for later reuse. A host test locks this behavior down.
- The Bambu zones editor remains intentionally canonical in the unified dashboard/API (`/api/v2/zones`) rather than being duplicated in provisioning.
- Handler budget: DragonBreath already supplies `httpd.max_uri_handlers = 48`; core copies that config and only raises undersized values. The observed real-hardware portal startup/routes also disproved the proposed 16-handler failure. Added a static contract so the 48-handler override cannot disappear silently.
- Favicon ordering: core invokes `register_product_routes` before registering `/*`; added an ordering contract to keep it that way.

New `db_portal_config` host coverage verifies credential-preserving None/unbind, unchanged form echoes, blank-secret retention, changed-field staging, and validation failure before persistence. All existing host suites, the new test, static API/portal checks, and a fresh ESP32-C3 compile/link at `3f4010f` pass locally.

This push has not yet been reflashed to hardware; the previous full HIL/real-unit battery covered the predecessor head and the shared-core stack fix. I am leaving the PR draft until CI and the focused portal hardware follow-up are complete.

plastikman (commented):
Re-review of `3f4010f` — the adapter findings are resolved, and I owe you a correction.

**Correction / retraction (my error):** my "handler-budget → `dc_portal_start` fails → no web server" finding was a **false positive**. `db_portal` sets `httpd.max_uri_handlers = 48` (it was there the whole time, at line 287 — I read the cfg struct just below it and missed the assignment) and `dc_portal` only *raises* undersized values, so it stays 48 — no 16-cap overflow. Your real-hardware run (portal up, all aliases + routes serving) is the proof. Sorry for the noise on that one; the static 48-handler contract you added is a good guard.

**Confirmed fixed:**
- ✅ **`/km-config` restored** — registered in `register_product_routes` (`km_config_get`), the Klipper-MQTT `/setup` section points to it, docs updated.
- ✅ **`apply_product` is now atomic** — parse→stage→validate all fields, then persist with `dc_source_set()` **last**, so the `ctl_src=1` + `ha_port=99999` case can't switch sources or partial-save.
- ✅ **Subsystem-specific backend error messages** (`Could not save %s settings: %s`).
- ✅ **Change detection** — an unchanged SPA echo writes nothing; blank secrets retain the saved value.
- ✅ **None-unbind retains creds** (per your decision) with a host test locking it.
- ✅ **Favicon ordering contract** (product routes before `/*`).
- ✅ Zones staying canonical in the dashboard/API rather than duplicated in provisioning — agreed.

From the DragonBreath adapter side this LGTM; the only thing outstanding is your own portal-hardware follow-up + CI on this head, which you're rightly holding the draft for.

justinh-rahb:
Exact-head CI is green on `3f4010f` (run 31293594875): all host tests, including the new portal validation/change-planning suite; HIL runner tests; cppcheck; normal ESP-IDF build; API/shared-dashboard contract checks; safe dev-board HIL-profile build; and firmware artifact upload all passed. The code-review items summarized above are resolved in the branch. Hardware has not yet been flashed with this new head, so the PR remains draft pending the focused portal regression pass on-device.

justinh-rahb:
@plastikman Progress update: the review follow-ups are now pushed at `3f4010f`. `/km-config` is restored, product configuration is fully validated before any write, unchanged form echoes no longer churn NVS, backend failures return useful messages, and blank secrets/source unbinding deliberately retain the stored credentials. Regression and contract coverage were added for those behaviors.

Exact-head CI is green across the host suites, HIL runner tests, cppcheck, normal ESP-IDF build, API/shared-dashboard checks, safe dev-board HIL-profile build, and artifact upload.

Next step is the focused on-device portal regression pass on this exact head: flash the app image while preserving the OEM bootloader/partition layout and existing configuration, verify startup plus `/`, `/setup`, `/km-config`, favicon and captive-portal routing, exercise unchanged and invalid saves to confirm no partial writes, and confirm credentials survive source changes. I’m leaving the PR draft until that hardware pass is complete.

justinh-rahb:
Repin + devboard HIL update on exact head `23c8b4e`:

- all eight shared components now pin the permanent dragon-core `v0.5.0` release
- the build guard now resolves annotated tags before comparing the Component Manager lock, so it rejects the stale `203164a` cache but accepts the correct `v0.5.0` tag-object SHA
- local HIL builds now go through that same guard, permanently covering the previously recurring missing-RISC-V-toolchain and stale-cache failures
- host/contract suites and the normal ESP-IDF build passed
- the local native-USB ESP32-C3 devboard was flashed with the safe compiled-out profile; all 3 scenarios / 66 steps passed, with the Panda GPIO/ADC exclusion audit also passing

The real DragonBreath was left untouched while it cools. PR remains draft pending the focused real-controller portal/OTA pass on this head.

justinh-rahb:
Exact-head GitHub CI is now green on `23c8b4e` (run 31294709801): every host suite, 13 HIL-runner tests, cppcheck, normal ESP-IDF build, API/shared-portal contract checks, both safe devboard HIL-profile builds, and artifact upload passed. Combined with the local native-USB physical HIL result (3 scenarios / 66 steps / 0 failures), the only remaining draft gate is the current-head real-controller portal/OTA regression after the unit has cooled.

plastikman (commented):
Re-review of `23c8b4e`: clean.

- **Repin to `dragon-core v0.5.0`** (stable tag) — confirmed v0.5.0 carries the hardened `dc_portal` (station-auth gate on provisioning + DNS self-delete teardown from the dc#5 merge). Good move going to a tag rather than a floating commit.
- **HIL builds routed through the guarded IDF preflight** (`idf-build.sh`/`hil.py`/`test_hil_runner.py`) — build tooling only.
- **No `db_portal.c`/`pb_httpd.c` changes**, so the adapter fixes verified at `3f4010f` (`/km-config`, atomic apply, per-subsystem errors, change detection, None-unbind, favicon ordering, the 48-handler override) are intact. CI green.

From the review side this looks ready to land, pending your portal-hardware pass on this exact head.

justinh-rahb:
Real-hardware follow-up on current head `96cac21`:

- Updated the supervised Panda controller through **app-only OTA** throughout; OEM bootloader, partition table, and NVS were never flashed or erased.
- Real-Panda non-heating HIL now passes cleanly: 1 scenario, 2/2 steps, with no `--allow-heater`. The controller remained OFF.
- The run exposed two issues that are now fixed on this branch:
  - inactive HA / Klipper-MQTT setters persisted NVS but left stale in-memory portal readback; getters also failed to load persisted config on the first request after reboot;
  - remote `--suite` staging could retain deleted/renamed scenarios from an earlier Pi run. Runner + scenarios are now isolated under a per-run root.
- Verified on the final release-profile image (`96cac21`) immediately after a clean boot:
  - first complete SPA echo returns `Configuration already up to date.` (no normalization/write);
  - invalid combined `ctl_src=1` + `ha_port=99999` returns HTTP 400 and leaves the source on Klipper;
  - blank secret fields retain saved credentials; control source and Moonraker configuration remain intact;
  - `/`, `/setup`, `/km-config`, and `/favicon.ico` all return 200 with the expected content types.
- Restored the controller to the exact-head **production** build via OTA. Final state: OFF, heater demand/output false, no fault, Moonraker connected.
- Host checks pass (`db_portal_config` and HIL runner 14/14). Earlier devboard HIL remains 3 scenarios / 66 steps / 0 failures.

Exact-head GitHub CI is still running, so I am leaving the PR draft until that completes.

plastikman (commented):
Re-review of `587d586` (`Keep inactive portal config snapshots current` + HIL run isolation):

**The fix is correct** — `db_klipper_mqtt_set_config` / `pb_ha_set_config` now update the in-memory `s_cfg` even when the component isn't started, so a set→readback within a session stays aligned. Good.

**One adjacent thing I want you to sanity-check (verified facts + an open question — not asserting a bug):**

Verified:
- `db_klipper_mqtt_get_config` / `pb_ha_get_config` (and core `dc_bambu`/`dc_moonraker`) return `s_cfg` **without** loading NVS; `s_cfg` is only populated on `start()` or `set_config`.
- On a **cold boot**, only the *active* source is started, so the other sources' `get_config()` return `{0}`.
- Both `describe_product()` and `apply_product()` base their values on `get_config()`.

Consequence I'm fairly sure of: on a fresh boot, `/setup` shows **blank fields for every non-active source** (the config is safe in NVS but not displayed) — a behavior change from the old `pb_portal`, which read NVS directly and showed all sources' saved values.

The open question is severity, and it lives in `db_portal_plan_product_save()` (which I couldn't fully trace): if a user **edits one field of a non-active source** on a cold boot and saves, is the whole struct written from that blank base — clobbering the source's other saved fields (host/user/inst/topic) — or does the plan preserve the untouched NVS values? Your host tests cover changed-field staging + blank-secret retention; do they cover the *empty-base + one edited field* case? If it can clobber, the clean fix is to have `get_config` lazy-load NVS when the component isn't started (or `apply_product` re-read NVS as the base), which also fixes the blank pre-fill.

Everything else here (HIL run isolation) is test infra. CI running on this head.

justinh-rahb:
Exact-head CI is now green on `96cac21` ([run 31299022899](https://github.com/plastikman/DragonBreath/actions/runs/31299022899)). With the full devboard + real-Panda validation complete and the production image restored safely, I have marked this ready for review.

justinh-rahb:
Confirmed — your consequence and severity concern are valid. On a cold boot, an inactive Bambu source supplied a zeroed base, so editing one Bambu field could write the whole staged struct and clobber untouched saved fields. Inactive Moonraker was worse: `dc_moonraker_get_config()` attempted to take its not-yet-created mutex.

The product-local HA/Klipper-MQTT half was fixed in `96cac21` (lazy persisted readback plus safe pre-start setters). I have deliberately **not** added a DragonBreath workaround for the core-backed sources. The owning fix is now in dragon-core PR [#7](https://github.com/justinh-rahb/dragon-core/pull/7): both Bambu and Moonraker config APIs lazy-load persisted values before startup, accept set-before-start safely, and clear their caches with NVS. Public lifecycle docs and the all-components compile/link app were updated as well.

Core host/UI tests and the full guarded ESP-IDF 5.3 ESP32-C3 build pass locally. I have returned #73 to draft; after #7 passes review/CI, I will merge it, tag a patch release, repin #73, and rerun the portal + hardware battery.

justinh-rahb:
Follow-up to the inactive-source review on current head `0b82dc1`:

- Confirmed the concern was real and fixed it at the owner: dragon-core [#7](https://github.com/justinh-rahb/dragon-core/pull/7) is merged and tagged `v0.5.1`. There is no DragonBreath workaround for core-backed Bambu/Moonraker config.
- #73 now pins **all eight** core components to `v0.5.1`; docs/changelog were updated accordingly.
- Added an app-side planning regression: a one-field inactive Bambu edit preserves its serial, access code, every unrelated source, and the selected source.
- Complete DragonBreath host suite passes; HIL runner 14/14; fresh guarded ESP-IDF build resolves all eight components to the single `v0.5.1` commit and passes the API/UI contract.
- Real Panda, app-only OTA (bootloader/partitions/NVS untouched):
  - exact-head non-heating HIL passes 2/2;
  - staged a reversible inactive-Bambu host-only edit with the access code omitted/retained;
  - after booting the exact-head production image, cold-boot `/setup` recovered that inactive value from NVS;
  - restored the original value, repeated the same save (`Configuration already up to date.`), rebooted again, and confirmed the original value remained restored;
  - Klipper source + Moonraker host/port remained intact throughout; hidden secret fields stayed redacted/retained.
- Final controller state: production `0b82dc1`, OFF, heater demand/output false, no fault, Klipper selected, Moonraker connected. `/`, `/setup`, `/km-config`, and `/favicon.ico` all return 200.

The local devboard is not currently enumerating, so its earlier unchanged-policy 66/66 result stands; this head received the full real-Panda/core-lifecycle pass. Waiting on exact-head CI before marking ready again.

plastikman (commented):
Re-review of `0b82dc1` — the inactive-source concern is fully resolved. Thanks for chasing it down.

- **App-side:** `db_klipper_mqtt_get_config` and `pb_ha_get_config` now lazy-load NVS when the component isn't started (`if (!s_lock) { nvs_load(&persisted); *out = persisted; }`), so provisioning describes the *persisted* config instead of a zeroed snapshot. That fixes both the blank `/setup` pre-fill and the clobber-on-edit risk (apply_product's base is now the real saved config). Reading NVS without the lock is safe here since an inactive component isn't touched by any other task.
- **Core-backed (Bambu/Moonraker):** same gap fixed at the owner in dragon-core #7 / `v0.5.1`; #73 repins all eight to `v0.5.1`. Good call fixing it in core rather than a local workaround.
- Host test (`db_portal_config_host_test.c`) + CHANGELOG updated; CI green.

From the review side #73 looks complete — adapter findings all resolved, pinned to the released `v0.5.1`. Over to your final hardware pass / merge.

justinh-rahb:
Needs a re-pin to core `v0.5.3`. This branch is currently at `v0.5.2` (head `babb15e`), and two follow-on dragon-core fixes have since landed — both found by hardware testing this PR.

**[dragon-core#8](https://github.com/justinh-rahb/dragon-core/pull/8) → `v0.5.2`** (already pinned here) — shared SPA provisioning/diagnostics repairs:

- provisioning **GETs** are now authenticated, not just mutations. `loadProvisioning()` was a bare `fetch('/api/v1/provisioning')`, so the setup plane failed outright against a token-protected DragonBreath. The new `dcFetchJson()` helper attaches `X-DragonBreath-Auth` and carries the existing 403 → re-prompt → retry flow onto reads.
- `/api/v1/system/logs` routes through that same helper instead of its own hand-rolled header/parse path.
- the full-screen setup plane and its cards are opaque again — `.provision-overlay` went from a 94% `color-mix` back to solid `var(--background)`, and `.provision-card` to an explicit `light-dark()` surface.
- firmware/device/boot identity renders at the top of `applyDeviceInfo()`, ahead of descriptor and capability gating, so an unknown `ui.schema` can no longer suppress firmware diagnostics.

**[dragon-core#9](https://github.com/justinh-rahb/dragon-core/pull/9) → `v0.5.3`** (the outstanding delta) — product-defined setup cards nest under `#dc-product-setup`, which had no layout spacing of its own; the static cards were getting their gap from `.provision-shell`. Adds the matching 10px column gap. Observed on real hardware after the `v0.5.2` OTA.

Both are locked into `tests/check_ui.mjs` (+25 lines across the two).

**Re-test scope:** `v0.5.1…v0.5.3` touches only `components/dc_ui/www/app.html` and `tests/check_ui.mjs` — no C surface, no component API, nothing in the safety path. So the repin wants a focused provisioning/setup pass on the controller (auth on setup load, log fetch, card spacing and opacity in both themes), not another full HIL cycle. The devboard and real-Panda results already recorded above still stand for everything else.

Action: bump all eight component pins to `v0.5.3`, refresh the lock, and confirm the build guard resolves the new tag object.

justinh-rahb:
Rebase, core `v0.5.4` repin, and real-controller qualification on exact head `f940052`.

## Rebased onto released `main`

`v1.1.0` shipped from `main` while this branch was open, which produced one conflict and one problem worth flagging.

- **Conflict:** [d7cb3b7](https://github.com/plastikman/DragonBreath/commit/d7cb3b7) edited the Bambu chamber-zone help text on `/setup`, a file this branch deletes. Resolved by keeping the deletion. `CHANGELOG.md` and `tools/package_release.sh` auto-merged, so the `dev_config.h` release guard from `d85f0d8` came across intact.
- **Changelog placement:** this branch had been appending its entries into `## [1.1.0]`, which is now a released section. Restored `1.1.0` verbatim from `main` (verified zero removals) and moved the portal-move, `dc_portal`, setup-UI and guarded-build entries under `[Unreleased]`.

## The zone note would have been silently reverted

`d7cb3b7`'s clarification had nowhere to land here: this branch removes zone editing from `/setup` entirely — the adapter declares only `bb_host` / `bb_serial` / `bb_code` — and the editors now live in the dashboard's **Filament zones** and **Custom profiles** cards. Those cards said only "while a Bambu print runs".

[dragon-core#10](https://github.com/justinh-rahb/dragon-core/pull/10) carries the wording upstream and locks it into `check_ui.mjs`, merged and tagged `v0.5.4`. All eight components are repinned from `v0.5.3` to `v0.5.4`. `v0.5.3…v0.5.4` is copy-only — `app.html` and `check_ui.mjs`, no C surface, no component API, nothing in the safety path.

`v0.5.3` was also retagged annotated so the whole `v0.5.x` series is uniform; the build guard resolves the tag object on both builds below.

## Hardware qualification

Supervised Panda controller, app-only OTA throughout. OEM bootloader, partition table and NVS were never written.

- Full 4 MB pre-flash backup, SHA256 `d8612ea1…`, held in two places.
- HIL-panda build at `f940052`, verified before writing: `PB_HIL_CONSOLE=y`, `PB_HIL_DEVBOARD` unset, real Panda GPIO/ADC backends live.
- Real-Panda non-heating HIL: **2/2 PASS**, no `--allow-heater`, controller OFF throughout.
- Production release build at `f940052` (`PB_POWER_LED=y`, HIL console absent), `0x112770`, **43% of each OTA app slot free**. Now running.
- Both OTAs echoed back a SHA byte-identical to the local image.
- Served SPA is **byte-identical to the `dc_ui` `v0.5.4` asset**, carrying the #8 auth/opacity/identity repairs, #9's card spacing, and #10's zone wording.
- Authenticated `GET /api/v1/provisioning` returns 20 networks and all five product sections with fields populated (Moonraker host reads back `10.0.3.207`); `GET /api/v1/system/logs` returns entries. Both 403 without the token, 200 with it — the path #8 repaired.
- Visual pass on the setup surface confirmed on hardware: opaque plane and cards, even spacing on product-defined cards, both themes, logs load.
- Build guard behaved correctly on both builds: quarantined the stale lock, re-resolved `v0.5.4`, and reported `Dependency lock matches every manifest git ref`.

Final state: mode `off`, heater output false, chamber 31.2 °C, PTC 30.9 °C, no fault, Moonraker connected.

Two notes for the record. The controller was reset once mid-session over RTS: a test-harness reverse proxy exhausted the LWIP socket table (`LWIP_MAX_SOCKETS=10`) and wedged `httpd_accept_conn`. That was the harness, not the firmware — `db_portal_start()` sets `lru_purge_enable`, the keep-alive tuning, and `stack_size 8192` exactly as `main` did. The safety loop was unaffected throughout: control ticks and zero-cross detection continued with `heater=off`. Separately, both OTA slots now hold `f940052` (release in app0, HIL in app1), so OTA rollback no longer reaches a pre-#73 image — the full-flash backup is the route back.

The Filament zones card only renders when the control source is Bambu, and this controller is bound to Klipper, so #10's wording is verified in the served bytes rather than on screen.

Exact-head CI green on `f940052`.

plastikman:
### Follow-up review @ `f940052`

Re-verification pass. The only delta since my last review (`9c7c08a`) is the dragon-core pin bump `v0.5.0 → v0.5.4` plus picking up `main` (release guard + changelog) — the DragonBreath-side product code is unchanged.

**Verified at this head:**

- **Pin coherence** — all 8 core components pinned consistently to `v0.5.4` in `main/idf_component.yml`; lock is CI-regenerated (no stale-lock risk).
- **Earlier findings stayed fixed:**
  - `/km-config` generator route registered (`register_product_routes`, `db_portal.c:524`).
  - Atomic apply — `apply_product` validates the whole plan first, then persists, with `dc_source_set` **last** (`db_portal.c:321-323`): no half-switched source if a sub-config write fails.
  - Inactive-source `get_config` lazy-loads persisted NVS for both `db_klipper_mqtt` and `pb_ha`, with a safe fallback to the zeroed static on NVS-miss. This is what keeps blank-secret retention correct when editing an inactive source.
- **Callbacks** — heater-off maintenance guard is conservative (`OFF && !heater_output`); `factory_reset` scoped to `app_nvs`; `/km-config` emits placeholders, never the stored password.
- **`validate_image`** accepting `panda_breath` alongside `dragonbreath` is the pre-existing intentional "revert to stock" behavior, faithfully ported — not a regression.
- **Host tests** — `db_portal_config_host_test` (27 assertions: unbind-retains-secrets, unchanged-echo-no-writes, blank-secret retention, inactive-source partial-edit retention) → PASS. The api-v2 contract check needs the build-time `dc_ui` SPA (absent in a host-only checkout; runs in CI post-build).

No new findings. Code looks merge-ready from a review standpoint — deferring to your focused real-controller portal/OTA regression pass on this head before it comes out of draft. Not approving ahead of that hardware gate.

PR #11, opened 2026-07-23
feat: add authoritative API v2

justinh-rahb:
## What

This is the second, stacked firmware change for the control rewrite. It builds on #9 and replaces the alpha HTTP contract with the authoritative DragonBreath API v2:

- removes `/status`, `/target`, `/heartbeat`, and `/reset`
- adds versioned info, state, command, heartbeat, health, and SSE event routes
- returns the complete authoritative state after accepted commands and most rejections
- makes control commands revision-aware while keeping OFF and drying-stop unconditionally safe
- requires the exact device-issued lease ID for POWER_ON heartbeats
- adds a small request-ID replay cache so a lost response can be retried without creating a second lease
- revision-guards fault clear inside `pb_policy`
- migrates the embedded dashboard to API v2, exact lease ownership, asynchronous SSE, and one-in-flight polling fallback
- adds the API contract document and CI checks for route removal, dashboard ownership, and embedded JavaScript syntax

## Why

The alpha API split state across ad-hoc endpoints and let clients infer ownership from their own requests. That made physical controls, another UI, a reconnect, or a slow/failed HTTP response difficult to reconcile safely. Polling also overlapped under slow Wi-Fi and contributed to the observed unresponsive UI.

API v2 makes the device snapshot authoritative. Clients consume the device's mode, source, revision, lease, output, sensors, environment, and fault state rather than assuming their last command still owns the heater.

## Protocol highlights

- `GET /api/v2/info`
- `GET /api/v2/state`
- `GET /api/v2/events` (SSE; max two asynchronous streams)
- `GET /api/v2/health`
- `POST /api/v2/command`
- `POST /api/v2/heartbeat`

The full contract and examples are in `docs/api-v2.md`.

The event task serializes the same snapshot used for its SSE ID, so an intervening state transition cannot produce an event whose ID and body disagree. Dashboard snapshots from the same boot are monotonic by revision; a new boot ID resets local ownership.

## Compatibility / impact

This is intentionally breaking alpha firmware. The existing `dragonbreath-klipper` helper will not control this branch until its separate API v2 migration lands. OTA and provisioning routes are preserved.

The dashboard and API v2 still need live hardware exercise before either stacked PR is merged.

## Checks

- `sh tests/run_policy_host_test.sh`
- `sh tests/check_api_v2_contract.sh`
- `node tests/check_dashboard_js.mjs`
- `git diff --check`
- full `espressif/idf:v5.3` ESP32-C3 build
[...]

PR #74, opened 2026-08-09
Restore /diag (product-local) + re-pin core v0.6.0 for /console

plastikman:
## Goal

The 1.1.1 portal extraction dropped the `/diag` and `/console` pages. Per the design discussion, they now have **different homes**, and this PR delivers the DragonBreath side of that split (v1.1.2).

- **`/diag` → product-local (`db_portal`).** Diagnostics are device-specific — chamber / element(PTC) / SSR duty / peak-element / fault-chain, over the read-only SSE stream (`/api/v2/events`), no USB. A generic capability-driven page would be a lukewarm lowest-common-denominator, so `/diag` stays device-specific and lives here.
- **`/console` → shared core.** Product-agnostic (raw `dc_evlog` ring), so it moved to `dragon-core` `dc_portal` (**justinh-rahb/dragon-core#11**, tagged **v0.6.0**). This PR picks it up via the re-pin.

## Changes

- Restore the device-specific `/diag` page in `db_portal` (registered before `dc_portal`'s `/*` SPA catch-all).
- Remove the temporary product-local `/console` — it now comes from core (avoids a double-registration).
- **Re-pin `dragon-core` v0.5.4 → v0.6.0**, which brings the core `/console` page plus the `dc_ui` setup-note fixes (Custom profiles AUTO-only note; "experts only" → web `/diag` instead of USB `tools/diag.py`).

## Validation
- Full ESP-IDF v5.3 ESP32-C3 build + link: PASS (`tools/idf-build.sh`); build guard reports "Dependency lock matches every manifest git ref," `v0.6.0` resolved for all components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,16 +7,26 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
-## [1.1.2] - 2026-08-09
+## [1.1.2] - 2026-08-10
+
+**Restores the `/diag` and `/console` pages dropped in 1.1.1, now split by where they
+belong.** Diagnostics are device-specific, so `/diag` is product-local; the firmware
+console is generic, so it moves to shared core.
 
 ### Fixed
-- **Restored the `/diag` and `/console` pages** (regressed in 1.1.1). The portal
-  extraction moved page ownership into `dragon-core` and these two product-local
-  diagnostic pages were dropped, leaving them 404 — their data endpoints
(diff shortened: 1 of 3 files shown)


justinh-rahb:
Should these not be additions to core?

plastikman:
I was trying to think about the same….

I think /diag is really breath specific, but /console probably makes a ton of sense in core. 

I can generalize the diag , it might be useful for the vent.

justinh-rahb:
PR on core is merged, will leave this one to you. Be advised some active cooking going on with core for Vent changes.

PR #89, opened 2026-08-24
bambu: add opt-in direct chamber temperature regulation

danielbrownjr:
## Summary

Adds optional direct chamber-temperature regulation for Bambu printers.

When enabled, DragonBreath uses the printer-reported Bambu chamber temperature as the regulation input instead of the local chamber NTC. The local chamber NTC and PTC remain independent safety inputs and retain authority over thermal limiting and fault handling.

The feature is opt-in and disabled by default.

## What changed

- expose Bambu chamber temperature and sample age through the DragonBreath state/API path
- allow fresh Bambu chamber telemetry to be used as the heater regulation input
- fall back to local chamber regulation when Bambu chamber telemetry is unavailable or stale
- retain the local chamber NTC as an independent thermal governor
- retain PTC monitoring and existing hard safety interlocks
- add Bambu chamber diagnostics to console/state output
- add a persisted `Use Bambu chamber sensor for heater control` setting under Bambu LAN configuration
- default direct Bambu chamber control to Disabled
- document the external regulation and local safety relationship

## Safety behavior

The Bambu chamber sensor is used only for temperature regulation.

It does not replace DragonBreath's local thermal safety sensors.

When direct chamber control is enabled:

- Bambu chamber temperature determines heater demand.
- The local chamber NTC independently limits heater output.
- The PTC sensor and existing fault/interlock paths remain active.
- Missing, invalid, or stale Bambu chamber telemetry cannot bypass local safety handling.

## Dependency update

This PR updates the DragonBreath dragon-core dependencies from `v0.27.0` to upstream `v0.28.2`.

During development, `dc_bambu` was temporarily pinned to my `feature/bambu-chamber-freshness` branch while the required stale chamber-telemetry handling was under review.

That change was merged upstream in `dragon-core#47` and released in `v0.28.0`. The submitted branch now uses `v0.28.2`, which retains the freshness support and also includes the `dragon-core#49` Bambu identifier log redaction. It no longer depends on my fork or an unreleased dragon-core commit.

## Testing

Tested on Panda Breath hardware connected to a Bambu X1C.

- ESP-IDF firmware builds successfully from a clean working tree.
- Hardware validation used upstream dragon-core `v0.28.0`; PR CI builds the submitted tree against `v0.28.2`.
- Direct chamber-control setting persists across reboot and firmware update.
- Fresh Bambu chamber telemetry is received and used for regulation when enabled.
- With a 60 C target and Bambu chamber telemetry around 39-40 C, the heater continued requesting heat above 60 C on the local NTC, confirming Bambu telemetry was driving regulation.
- The independent local chamber governor repeatedly disabled heater output around 72 C and re-enabled it after cooling to roughly 67 C.
- PTC monitoring remained active and healthy.
- No heater faults were observed during the validation runs.
- Disabling direct chamber control preserves the existing local-NTC regulation behavior.

## Related

- `justinh-rahb/dragon-core#47` - stale Bambu chamber-temperature expiry support
- `justinh-rahb/dragon-core#49` - follow-up log redaction for Bambu device identifiers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ sdkconfig.old
 managed_components/
 dependencies.lock
 
+# Local firmware checkpoints
+firmware-checkpoints/
+
 # Editor / OS
 .vscode/
 .idea/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,23 @@ below into the GitHub Release notes.
   disengages — and an incomplete/partial status response is treated as loss-of-source
   rather than heating on a stale bed target. Validated end-to-end by a Prusa user.
   Bumps **dragon-core** to **v0.28.1**.
+- **Bambu printer-chamber diagnostics.** `GET /api/v2/state` now exposes
+  `environment.printer_chamber_temperature_c` and `printer_chamber_age_ms`.
+  Stale or unavailable Bambu chamber samples are reported as `null`.
 
 ### Fixed
 - **Bambu chamber temperature no longer goes stale silently** — a chamber reading that
   stops updating over MQTT is now expired rather than held indefinitely (dragon-core #47).
+
+### Changed
+- **Bambu chamber regulation now follows the printer's chamber sensor** while keeping
+  DragonBreath's local chamber NTC and PTC safety-authoritative. External-sensor
+  regulation is bounded by a local chamber thermal limiter: heat cuts at **72 °C**
+  local and may resume below **67 °C**, while the independent hard chamber
+  over-temperature latch remains unchanged.
+- **Privacy:** `environment.bambu_serial` is no longer exposed by API v2 state
+  responses. The configured printer serial remains internal and is still used for
+  Bambu MQTT topic selection.
 
 ## [1.1.11] - 2026-08-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,17 @@ below into the GitHub Release notes.
   DragonBreath's local chamber NTC and PTC safety-authoritative. External-sensor
   regulation is bounded by a local chamber thermal limiter: heat cuts at **72 °C**
   local and may resume below **67 °C**, while the independent hard chamber
-  over-temperature latch remains unchanged.
+  over-temperature latch remains unchanged. Direct printer-chamber regulation applies
+  only in **Auto**; **Manual** and **Drying** always regulate from DragonBreath's local
+  chamber NTC.
+- **External-regulation operating range:** hardware validation was performed at a
+  **60 °C** chamber target. Because the local soft limiter cuts at 72 °C, targets near
+  the configurable 65–70 °C ceiling may be limited before the printer's bulk chamber
+  sensor reaches setpoint, depending on heater/outlet temperature rise.
+- **Bambu model compatibility:** direct chamber regulation currently depends on the
+  legacy `chamber_temper` telemetry field. Bambu models that report chamber temperature
+  only through newer `device.ctc.info.temp` telemetry will safely fall back to local
+  NTC regulation until that source is supported by `dc_bambu`.
 - **Privacy:** `environment.bambu_serial` is no longer exposed by API v2 state
   responses. The configured printer serial remains internal and is still used for
   Bambu MQTT topic selection.
--- a/components/db_portal/db_portal.c
+++ b/components/db_portal/db_portal.c
@@ -22,6 +22,31 @@
 
 static const char *TAG = "db_portal";
 
+#define DB_NVS_NAMESPACE "app_nvs"
+#define DB_NVS_KEY_BAMBU_CHAMBER_CTL "bb_ch_ctl"
+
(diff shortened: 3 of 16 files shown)


justinh-rahb (approved):
Reviewed the resolved branch against current main and dragon-core v0.28.2. The external chamber input remains regulation-only; the local chamber NTC, PTC limits, fault handling, stale-sample fallback, and opt-in default remain intact. Conflict resolutions preserve the Prusa planner and source capability API. All host test suites and diff checks pass. Approving, with final merge gated on the v0.28.2 firmware CI run.

justinh-rahb:
@plastikman I resolved the stale-main conflicts by merging current main into the contributor branch, preserved the existing Prusa and portal changes, and updated all dragon-core pins to v0.28.2. We will squash-merge this PR once the v0.28.2 firmware CI run is approved and green, so the contributor's development commits and the main merge commit land as one clean feature commit.

plastikman (approved):

plastikman:
Did an independent pass focused on the safety model — thanks for the thorough write-up and hardware validation. **The safety architecture is sound**: the external chamber temperature only ever drives the bang-bang regulation input (`regulation_c`); it can never bypass a cutoff. Every trip still reads the local sensors on unchanged paths (sensor-fault fail-closed, the 85 °C hard chamber trip, PTC 105 °C + element foldback), and the new 72/67 °C `s_local_cut` is active only under external regulation, fails safe on a bad local read, and is test-asserted to sit below the 85 °C trip. Worst-case a stuck-**low** printer reading (the dangerous direction) is capped by the local 72 → 85 → 105 governors — no runaway; a stuck-high reading just stops heat. Nice touch re-pushing the control temp every policy tick so a stale/NAN sample falls back to the local NTC with no latch. The boundary + hysteresis tests plus the `CUT_C < CHAMBER_MAX_C` / `RESUME < CUT` invariant asserts are exactly right.

A few non-blocking notes:

1. **Regulation isn't gated by heater mode.** `control_chamber_c` is pushed unconditionally each tick (gated only on `mk_connected && isfinite`), so with the opt-in on and Bambu connected+fresh, the printer chamber temp drives regulation in **Manual** and **Drying** too, not just Auto. Drying is the awkward one: a dry cycle (say 60 °C) with the printer idle-but-connected would regulate to the printer's ~ambient chamber reading and keep heating past the drying setpoint on the local sensor — stopping only at the 72 °C soft-cut instead of holding 60. Not unsafe (the local cap holds), but it wouldn't hold the drying target. Might be worth scoping to Auto (or to `bs.printing`), or documenting the mode scope.

2. **Setpoint headroom.** With the settable ceiling ~70 °C and the local soft-cut at 72 °C, there's only ~2 °C of local headroom — a high external target (65–70) could throttle at the local cut before bulk air reaches setpoint. Validated at 60, which is the common case; worth a doc note on the practical external-regulation ceiling.

3. **Newer Bambu models.** `dc_bambu` reads the legacy flat `chamber_temper`; H2/newer moved chamber temp to `device.ctc.info.temp`, so `chamber_temp` is absent → NAN → the feature silently no-ops (safely) on those printers. Worth calling out so owners aren't surprised.

4. Minor: the `app_nvs` / `bb_ch_ctl` namespace+key constants are duplicated in `app_main.c` and `db_portal.c` — a shared header would prevent drift. And `dc_ui` (line ~2073) still references `environment.bambu_serial`, which this PR intentionally stops emitting — it degrades gracefully to "Bambu", just a dead reference to tidy later.

None of these block it — the local-sensor safety authority is preserved throughout. 👍

justinh-rahb:
@plastikman Good point. I do not think we should add the H2/newer `device.ctc.info.temp` path for this feature: those models already have their own chamber heaters and should retain their native chamber control. This change is effectively aimed at the P2S, where Breath can supply the missing chamber heat while regulating from the printer-reported chamber temperature. The U1 could technically use the same mechanism, but its chamber sensor is inaccurate enough that I would not present it as a recommended target. So the H2 no-op is better treated as outside the intended applicability than as a compatibility gap; we should scope the docs/UI accordingly.

plastikman:
Followed the CI red→green on the merge-conflict resolution — the resolution looks right to me.

Scoping the external chamber reading to `PB_MODE_AUTO` (5cf8442) is the safety-positive call: Manual and Drying are local device modes and should keep regulating from the DragonBreath chamber NTC even when Bambu telemetry is connected and fresh. The forwarding seam only substitutes the printer-reported chamber temp where the feature is actually meant to apply (AUTO follow); everywhere else `control_chamber_c` stays NAN and pb_heater falls back to the local sensor.

`test_external_chamber_control_forwarding` now correctly drives AUTO before asserting the value reaches the heater seam (6b3f398), and the added "Manual regulation always stays on the local chamber NTC" case locks in that the guard is intentional rather than incidental. Nice to have that invariant pinned by a test.

Also good on scoping this to the P2S and deliberately not wiring the H2/newer `device.ctc.info.temp` path — those models have native chamber heaters and should keep native control.

No blocking issues from me; safety model still reads sound (local sensors retain all veto authority; external temp is a regulation input only).

PR #90, opened 2026-08-24
RFC: define Dragon device integration planes

justinh-rahb:
## Summary

Starts a numbered RFC series for the emerging Dragon device architecture:

- RFC 0001 defines the umbrella model and cross-plane invariants
- RFC 0002 scopes the wired Klipper-MCU safety spine for heater-class devices
- RFC 0003 scopes both the existing reliable klipper-esp32 ESP-NOW MCU stream and the proposed low-power sleeping-leaf fabric
- RFC 0004 defines semantic Dragon peer capabilities and remote-sensor regulation
- adds RFC status/review conventions and a discoverable plans index

All four RFCs are Proposed. Merging this PR records the proposals for discussion; it does not mark them Accepted or commit us to implementation.

## Feedback requested

The documents call out detailed questions, but the highest-value review is:

1. Is the three-plane split correct for known Dragon products and accessories?
2. Should generic remote regulation suspend by default when its sensor becomes stale, or fall back locally?
3. Should the existing reliable klipper-esp32 ESP-NOW stream remain the continuously connected MCU path while sleeping leaves use a smaller Dragon capability protocol, and can both safely share the same frame envelope?
4. Is leaf-to-gateway the right initial topology?
5. Which pairing and identity concepts should be shared across wireless leaves and smart peers?
6. Which first prototype would test the architecture with the least hardware and safety risk?

## Grounding and validation

- incorporates the working [`klipper-esp32` `codex/transport-abstraction` branch](https://github.com/justinh-rahb/klipper-esp32/tree/codex/transport-abstraction) at [`bfb5d9b`](https://github.com/justinh-rahb/klipper-esp32/commit/bfb5d9baad3350b3a18d05e73e3334a871c5dade): S3 USB bridge to a C3 Klipper MCU over reliable ESP-NOW
- distinguishes that continuously connected prototype from unimplemented sleeping/battery-leaf requirements such as pairing, authentication, multi-node operation, deep sleep, and measured power budgets
- preserves DragonBreath product-local safety authority and dragon-core boundaries
- `git diff --check` passes
- relative links were checked against the current repository tree
- documentation only; no firmware behavior changes

### DIFF
--- a/plans/README.md
+++ b/plans/README.md
@@ -1,0 +1,19 @@
+# DragonBreath plans and RFCs
+
+The `plans/` directory contains design records for DragonBreath and the wider Dragon
+firmware family.
+
+New cross-product proposals use the numbered [Dragon RFC series](rfcs/README.md).
+They are merged as discussion records before implementation and retain an explicit
+status throughout their lifecycle.
+
+## Current proposed RFC series
+
+- [RFC 0001: Dragon device integration planes](rfcs/0001-dragon-integration-planes.md)
+- [RFC 0002: Wired safety spine for heater-class devices](rfcs/0002-wired-safety-spine.md)
+- [RFC 0003: ESP-NOW low-power peripheral fabric](rfcs/0003-esp-now-peripheral-fabric.md)
+- [RFC 0004: Dragon peer capabilities and remote sensors](rfcs/0004-peer-capability-plane.md)
+
+The other files in this directory are earlier product RFCs, implementation designs,
+and retained architecture records. Their existing locations and status text remain
+authoritative; they do not need to be renumbered to participate in the new process.
--- a/plans/rfcs/0001-dragon-integration-planes.md
+++ b/plans/rfcs/0001-dragon-integration-planes.md
@@ -1,0 +1,182 @@
+# RFC 0001: Dragon device integration planes
+
+Status: **Proposed — feedback requested.**
+
+Date: 2026-08-24
+
+## Summary
+
+The Dragon family should use three distinct integration planes, selected by device
+class rather than by a single preferred transport:
+
+1. a **wired safety spine** for mains-powered, actuator-bearing, timing-sensitive
+   hardware;
+2. an **ESP-NOW peripheral fabric** for simple, low-power leaf devices; and
+3. a **Dragon peer capability plane** for powered, intelligent products exchanging
+   semantic state and capabilities.
+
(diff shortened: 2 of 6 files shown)


plastikman (commented):
Read through all four RFCs — strong, coherent series. The most useful thing I can add is that RFC 0001's cross-plane invariants aren't speculative; three of them describe firmware we just shipped/reviewed, which is good grounding for the "Proposed" state:

- **Invariant #3** ("freshness is decided by the consumer; track local receipt time; a remote clock can't prove a sample is fresh") is exactly dragon-core #51 (Prusa freshness — consumer-side expiry on `esp_timer` receipt time, provider timestamps ignored).
- **Invariant #1** ("actuator owner retains safety authority; a remote input must not disable local cutoffs/watchdogs") is exactly DragonBreath #89 (external Bambu chamber temp is a regulation input only; local NTC/PTC keep veto authority).
- **Invariant #5** ("loss behavior is part of configuration; indefinite stale is never valid") is the fail-cold posture in both.

**On the suspend-vs-fallback default (0004 / Q2):** endorse `suspend` as the generic default with `fallback_local` opt-in. The two live precedents map cleanly onto that per-capability choice: Prusa (#51) has no local equivalent → suspend/fail-cold; Bambu chamber (#89) has a local NTC → `fallback_local` (opt-in, off by default). So the rule "suspend by default, fall back only where a trustworthy local sensor exists" is already how the code behaves.

**One data point for 0001 (updated after the LCD sniffer shipped):** DragonWheeze is a clean example of the "one ESP32 = complete product" fourth shape, and an interesting *split*: it **actuates open-loop but observes its own state**. It drives the dryer with blind optocoupler button presses (no per-command acknowledgement that a press registered), yet it now **passively sniffs the panel's TM1621 LCD to read the dryer's actual temperature, humidity, and time** — so it can verify the resulting state even when it can't confirm an individual press landed. That distinction — *un-acknowledged actuation* vs. *observable outcome* — is worth naming distinctly from "stale remote sensor": invariant #3's "consumer owns validity/freshness" applies to the sniffed readback (freshness by local receipt time), while the blind actuation is a separate property. (This corrects my earlier note, which predated the sniffer and wrongly said the wheeze can't read the panel.)

Docs-only and all Proposed, so no objection to merging to record the proposals — the architecture is coherent and the safety invariants are proven by shipping code.

PR #91, opened 2026-08-30
control: port chamber PID to dragon-core dc_pid

danielbrownjr:
## Summary

Ports DragonBreath chamber control to the shared `dragon-core` `dc_pid` controller while preserving DragonBreath's product-owned heater policy and safety layers.

This PR is a downstream integration consumer of the `dc_pid` work merged in `justinh-rahb/dragon-core#54`.

## What changed

- Replace the private chamber PID backend with shared `dc_pid`.
- Use the hardware-selected Candidate A parameters:
  - Kp `0.100`
  - Ki `0.001`
  - Kd `0.040`
  - derivative alpha `0.20`
- Preserve the 500 ms controller cadence and 10-second SSR time-proportioning window.
- Apply DragonBreath-owned approach ceilings:
  - positive error below 2 C: 40%
  - positive error below 5 C: 70%
  - otherwise: 100%
- Expose the active approach ceiling to `dc_pid` and contract stored integral demand as that ceiling tightens, preventing hidden integral through the 100% -> 70% -> 40% trajectory.
- Force heater output off at/above target without discarding valid controller history.
- Preserve current temperature-source policy:
  - AUTO may use eligible fresh Bambu telemetry.
  - stale/missing/ineligible Bambu telemetry falls back to local NTC.
  - Manual and Drying remain local-NTC controlled.
- Reset controller history when the effective temperature source changes.
- Hold PID integration while local chamber or PTC foldback inhibits output.
- Leave existing local sensors, hard trips, foldbacks, watchdog, interlocks, target policy, fan behavior, and the single SSR writer authoritative and unchanged.
- Reject zero-cross edges arriving less than 4,000 us after the last accepted edge. Rejected edges do not advance accepted count/timing or trigger a fan transition; a separate rejected-edge counter is exposed to HIL and runtime logs.

## dragon-core dependency

All DragonBreath dragon-core components, including `dc_pid`, use the upstream release:

`justinh-rahb/dragon-core@v0.32.0`

The focused PID host runner uses the same upstream version as the production manifest. The temporary contributor-fork/SHA dependency has been removed.

## Code and CI validation

- 10/10 C host-test runners pass.
- Focused PID coverage validates source selection/fallback, source-transition reset, Candidate A gains, approach ceilings, the continuous 100% -> 70% -> 40% cap transition, SSR windowing, at-target heater cutoff, integration hold, and local foldback dominance.
- Zero-cross host coverage validates rejection of the observed approximately 1 ms duplicate edge, preservation of accepted timing across a rejection, valid 50/60 Hz intervals, and the exact 4,000 us threshold boundary.
- HIL runner unit suite: 19/19 pass.
- Panda smoke now keeps heater/fan off while checking accepted-edge cadence over five seconds, accepted interval range, and rejected-edge-counter observability.
- API v2 contract checks pass.
- Production ESP32-C3 build passes with first-party warnings treated as errors.
- Native-USB safe-HIL, UART safe-HIL, and native-USB production-network debug builds pass.
- GPIO/ADC compile-out verification passes for all safe dev-board profiles.
- Focused cppcheck over changed C sources passes.

Local cppcheck 2.19 reports an unchanged GNU `asm` parsing error in `db_portal.c`; this PR intentionally does not modify unrelated portal code.

## Hardware status

- Generic `dc_pid` correctness is validated by its upstream component tests and this product integration suite.
- DragonBreath product tuning adopted Candidate A from the real-Panda A/B results already recorded in this PR.
- Duplicate-edge ZCD handling is now fixed in software and protected by host/HIL coverage.
- Remaining hardware gate: run the non-heating Panda smoke scenario on a real Panda and verify approximately one accepted zero-cross per half-cycle (about 120/s on 60 Hz), accepted intervals near 8.33 ms, rejected duplicate/noise edges observable, and heater/fan remaining off.

This PR is not being represented as fully hardware-validated until that physical ZCD cadence requalification passes.

## Notes

The numerical behavior is not bit-for-bit identical to the private `pb_pid` implementation. `dc_pid` uses its documented derivative-on-measurement and anti-windup behavior.

This PR intentionally does not carry forward the old branch's private/selectable PID backends, controller-selection UI, experimental settings UX, PID diagnostics additions, HTTP-memory changes, SSE instrumentation, fallback experiments, or unrelated portal/source-policy changes.

### DIFF
--- a/.github/workflows/firmware-build.yml
+++ b/.github/workflows/firmware-build.yml
@@ -63,6 +63,9 @@ jobs:
       - name: Test heater safety-trip decision (over-temp / fail-closed / watchdog)
         run: sh tests/run_heater_safety_host_test.sh
 
+      - name: Test chamber dc_pid integration and safety hold
+        run: sh tests/run_heater_pid_host_test.sh
+
       - name: Test MQTT-Klipper arming state machine (retained-aware arm/heartbeat)
         run: sh tests/run_db_klipper_mqtt_host_test.sh
 
--- a/.github/workflows/firmware-build.yml
+++ b/.github/workflows/firmware-build.yml
@@ -65,6 +65,9 @@ jobs:
       - name: Test heater safety-trip decision (over-temp / fail-closed / watchdog)
         run: sh tests/run_heater_safety_host_test.sh
 
+      - name: Test zero-cross duplicate-edge filter
+        run: sh tests/run_fan_zc_host_test.sh
+
       - name: Test chamber dc_pid integration and safety hold
         run: sh tests/run_heater_pid_host_test.sh
 
--- a/components/pb_fan/CMakeLists.txt
+++ b/components/pb_fan/CMakeLists.txt
@@ -1,5 +1,5 @@
 idf_component_register(
-    SRCS "pb_fan.c"
+    SRCS "pb_fan.c" "pb_fan_zc_filter.c"
     INCLUDE_DIRS "include"
     REQUIRES pb_board driver esp_timer
 )
--- a/components/pb_fan/include/pb_fan.h
+++ b/components/pb_fan/include/pb_fan.h
@@ -19,10 +19,11 @@ esp_err_t pb_fan_init(void);
 void pb_fan_set_level(uint8_t percent);
 uint8_t pb_fan_get_level(void);
 
-// Diagnostics: total zero-cross interrupts since boot, and microseconds between
-// the last two edges (mains cycle period). interval 0 = no zero-cross seen yet
-// (ZCD not working / no mains) — lets us verify the detector without firing.
-void pb_fan_zc_diag(uint32_t *count_out, uint32_t *interval_us_out);
+// Diagnostics: accepted edges since boot, interval between the last two accepted
+// edges, and rejected implausibly-close edges. interval 0 = fewer than two
+// accepted edges (ZCD not working / no mains).
+void pb_fan_zc_diag(uint32_t *count_out, uint32_t *interval_us_out,
+                    uint32_t *rejected_count_out);
 
 #ifdef CONFIG_PB_DEVBOARD_SAFE
 // Inject one or more zero-cross events into the dev-board backend.
--- a/components/pb_fan/include/pb_fan_zc_filter.h
+++ b/components/pb_fan/include/pb_fan_zc_filter.h
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+// Pure zero-cross edge filter shared by the Panda ISR and host tests.
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+// Reject the ~1 ms duplicate edges observed on Panda while retaining valid
+// 60 Hz (~8.33 ms) and 50 Hz (10 ms) half-cycle edges.
+#define PB_FAN_ZC_MIN_INTERVAL_US 4000U
+
+typedef struct {
+    volatile uint32_t accepted_count;
+    volatile uint32_t accepted_interval_us;
+    volatile uint32_t rejected_count;
+    volatile uint64_t last_accepted_us;
+} pb_fan_zc_filter_t;
+
+void pb_fan_zc_filter_reset(pb_fan_zc_filter_t *state);
+
+// Record one observed edge. Rejected edges increment only rejected_count; they
+// never advance accepted timing/count state.
+bool pb_fan_zc_filter_record(pb_fan_zc_filter_t *state, uint64_t now_us);
--- a/components/pb_fan/pb_fan.c
+++ b/components/pb_fan/pb_fan.c
@@ -11,6 +11,7 @@
 // ONE zero-cross-synced pulse per half-cycle (phase control) — never PWM.
 #include "pb_fan.h"
 #include "pb_board.h"
+#include "pb_fan_zc_filter.h"
 
 #include "esp_log.h"
 #include "esp_timer.h"
@@ -21,17 +22,16 @@ static const char *TAG = "pb_fan";
 
 static volatile bool     s_want_on;
 static volatile bool     s_applied;
-static volatile uint32_t s_zc_count;
-static volatile uint32_t s_zc_interval_us;
-static volatile uint64_t s_zc_last_us;
+static pb_fan_zc_filter_t s_zc;
+#ifdef CONFIG_PB_DEVBOARD_SAFE
+static uint64_t s_hil_now_us;
+#endif
 
 #ifndef CONFIG_PB_DEVBOARD_SAFE
 static void IRAM_ATTR zcd_isr(void *arg)
 {
     uint64_t now = esp_timer_get_time();
-    if (s_zc_last_us) s_zc_interval_us = (uint32_t)(now - s_zc_last_us);
-    s_zc_last_us = now;
-    s_zc_count++;
+    if (!pb_fan_zc_filter_record(&s_zc, now)) return;
 
     bool on = s_want_on;                 // apply the commanded level at the zero cross
     if (on != s_applied) {
@@ -46,12 +46,15 @@ esp_err_t pb_fan_init(void)
 #ifdef CONFIG_PB_DEVBOARD_SAFE
     s_want_on = false;
     s_applied = false;
-    s_zc_count = 0;
-    s_zc_interval_us = 0;
-    s_zc_last_us = 0;
+    pb_fan_zc_filter_reset(&s_zc);
+    s_hil_now_us = 0;
     ESP_LOGW(TAG, "safe dev-board backend: fan gate and ZCD GPIOs compiled out");
     return ESP_OK;
 #else
+    s_want_on = false;
+    s_applied = false;
+    pb_fan_zc_filter_reset(&s_zc);
+
     const gpio_config_t gate = {
         .pin_bit_mask = (1ULL << PB_GPIO_FAN_GATE),
         .mode = GPIO_MODE_OUTPUT,
@@ -75,8 +78,6 @@ esp_err_t pb_fan_init(void)
     esp_err_t err = gpio_isr_handler_add(PB_GPIO_ZERO_CROSS, zcd_isr, NULL);
     if (err != ESP_OK) return err;
 
-    s_want_on = false;
-    s_applied = false;
     ESP_LOGI(TAG, "init: ON/OFF held-gate (stock model), gate GPIO%d, ZCD GPIO%d; never PWM'd",
              PB_GPIO_FAN_GATE, PB_GPIO_ZERO_CROSS);
     return ESP_OK;
@@ -98,18 +99,25 @@ void pb_fan_set_level(uint8_t percent)
 
 uint8_t pb_fan_get_level(void) { return s_want_on ? 100 : 0; }
 
-void pb_fan_zc_diag(uint32_t *count_out, uint32_t *interval_us_out)
+void pb_fan_zc_diag(uint32_t *count_out, uint32_t *interval_us_out,
+                    uint32_t *rejected_count_out)
 {
-    if (count_out) *count_out = s_zc_count;
-    if (interval_us_out) *interval_us_out = s_zc_interval_us;
+    if (count_out) *count_out = s_zc.accepted_count;
+    if (interval_us_out) *interval_us_out = s_zc.accepted_interval_us;
+    if (rejected_count_out) *rejected_count_out = s_zc.rejected_count;
 }
 
 #ifdef CONFIG_PB_DEVBOARD_SAFE
 void pb_fan_hil_zero_cross(uint32_t count, uint32_t interval_us)
 {
     if (count == 0) return;
-    s_zc_count += count;
-    s_zc_interval_us = interval_us;
-    s_applied = s_want_on;
+    for (uint32_t i = 0; i < count; i++) {
+        if (s_hil_now_us == 0)
+            s_hil_now_us = 1;
+        else
+            s_hil_now_us += interval_us;
+        if (pb_fan_zc_filter_record(&s_zc, s_hil_now_us))
+            s_applied = s_want_on;
+    }
 }
 #endif
--- a/components/pb_fan/pb_fan_zc_filter.c
+++ b/components/pb_fan/pb_fan_zc_filter.c
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: MIT
+#include "pb_fan_zc_filter.h"
+
+#ifdef ESP_PLATFORM
+#include "esp_attr.h"
+#else
+#define IRAM_ATTR
+#endif
+
+#include <limits.h>
+
+void pb_fan_zc_filter_reset(pb_fan_zc_filter_t *state)
+{
+    if (!state) return;
+    state->accepted_count = 0;
+    state->accepted_interval_us = 0;
+    state->rejected_count = 0;
+    state->last_accepted_us = 0;
+}
+
+bool IRAM_ATTR pb_fan_zc_filter_record(pb_fan_zc_filter_t *state,
+                                       uint64_t now_us)
+{
+    if (!state) return false;
+
+    const uint64_t last = state->last_accepted_us;
+    if (last != 0) {
+        if (now_us <= last || now_us - last < PB_FAN_ZC_MIN_INTERVAL_US) {
+            state->rejected_count++;
+            return false;
+        }
+
+        const uint64_t interval = now_us - last;
+        state->accepted_interval_us = interval > UINT32_MAX
+            ? UINT32_MAX : (uint32_t)interval;
+    }
+
+    state->last_accepted_us = now_us;
+    state->accepted_count++;
+    return true;
+}
--- a/components/pb_heater/CMakeLists.txt
+++ b/components/pb_heater/CMakeLists.txt
@@ -1,5 +1,5 @@
 idf_component_register(
     SRCS "pb_heater.c"
     INCLUDE_DIRS "include"
-    REQUIRES pb_board pb_ntc driver esp_timer nvs_flash
+    REQUIRES pb_board pb_ntc dc_pid driver esp_timer nvs_flash
 )
--- a/components/pb_heater/include/pb_heater.h
+++ b/components/pb_heater/include/pb_heater.h
@@ -176,7 +176,7 @@ static inline bool pb_heater_foldback_cut(bool ptc_ok, float ptc_c, bool prev_cu
 // priority ordering can be host-tested without the ADC/SSR/RTOS backend. Given the
 // freshest per-channel sensor reads (status-OK flags + instantaneous °C), whether a
 // target is armed, and whether the comms deadman has expired, returns the fault the
-// tick must latch — or PB_FAULT_NONE if it is safe to run the bang-bang loop. The
+// tick must latch — or PB_FAULT_NONE if it is safe to run the control loop. The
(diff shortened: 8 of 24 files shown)


justinh-rahb (changes requested):
One control-correctness issue remains.

[P2] Make the approach cap visible to PID anti-windup — components/pb_heater/include/pb_heater_pid.h:85-104

dc_pid is configured with output_max = 1.0, then its output is capped to 0.40 or 0.70 afterward. Because the controller cannot see this lower actuator ceiling, it continues accumulating integral while delivered duty is capped. I reproduced this at a constant 4 °C error: after 2,000 samples, the stored integral reached approximately 0.90 and raw P+I demand approximately 1.0 while delivered duty remained 0.40. That leaves demand far above the effective ceiling and can produce prolonged near-target cycling or overshoot.

Please feed the active approach limit into the controller saturation policy, or otherwise prevent accumulation when the approach cap truncates output, and add a sustained capped-output regression test.

The safety/foldback ownership, source-transition reset, and SSR windowing otherwise look coherent. The focused PID integration test and current GitHub build pass. The contributor-fork/SHA dependency remains the already-documented draft merge blocker now that dragon-core #54 has merged.

danielbrownjr:
Implemented the requested control-correctness follow-up in `eb21c9f`.

What changed:
- DragonBreath now computes its existing 0.40 / 0.70 / 1.00 approach ceiling before each PID step and passes that active ceiling as `dc_pid.output_max`.
- `dc_pid` therefore applies its existing conditional anti-windup against the duty the actuator can actually receive; the old post-PID truncation is removed.
- At/above target, DragonBreath still commands zero while stepping through a valid controller range so measurement/derivative history and negative-error integral unwinding remain coherent.
- Product safety/fault foldbacks remain downstream and authoritative. Their existing `integrate=false` hold path is unchanged.
- The temporary contributor-fork pin was removed. `dc_pid` now comes from `https://github.com/justinh-rahb/dragon-core.git` tag `v0.31.0`, which contains merged PR #54.

Regression coverage:
- 4 C fixed error for 4,000 samples under the 0.40 ceiling: delivered duty stays at/below 0.40 and P+I remains at/below the effective ceiling instead of winding toward 1.0;
- cap relaxation from 0.40 to 0.70: stored integral changes by only the next normal integration increment, with no hidden-demand jump;
- 8 C fixed error for 4,000 samples under the 0.70 ceiling;
- 12 C fixed error for 4,000 samples with the unchanged 1.00 ceiling;
- existing local foldback/integration-hold, source reset/fallback, heater-only cutoff, and SSR-window tests remain passing.

Validation:
- all nine C host runners pass;
- HIL runner unit suite: 14/14 pass;
- API v2/shared-dashboard contract check passes;
- focused cppcheck passes;
- fresh ESP-IDF 5.3.1 ESP32-C3 production build passes with first-party `-Wall -Wextra -Werror`;
- safe dev-board and UART HIL builds pass;
- GPIO/ADC compile-out checks pass for both HIL profiles.

No hardware heating test was performed.

justinh-rahb (changes requested):
The fixed-error anti-windup case is improved, but one blocking transition case remains.

[P2] Tightening the approach cap preserves demand above the new ceiling — components/pb_heater/include/pb_heater_pid.h:85-103

Passing approach_cap as output_max prevents new accumulation when the controller starts inside that band, but it does not unwind integral accumulated under a wider band. I reproduced the realistic transition from 12 °C error to 4 °C error: after the 100% band, integral was approximately 0.6984; after 2,000 further samples under the 40% cap, it remained 0.6984, leaving raw P+I demand around 0.7984 while delivered duty stayed at 0.4. The new tests reset state between bands and only exercise cap relaxation, so they miss the 100%→70%→40% trajectory the heater actually follows. Please normalize or back-calculate stored demand when the active ceiling tightens and add downward-transition coverage.

[P3] Host test still uses the obsolete fork dependency — tests/run_heater_pid_host_test.sh:8-14

The firmware manifest now consumes justinh-rahb/dragon-core v0.31.0, but this runner still downloads commit 3447fd4 from danielbrownjr/dragon-core. The behavioral host test therefore does not compile the production dependency and still relies on the contributor fork. Please update it to fetch the upstream release or otherwise derive the tested source from the manifest.

The upstream firmware dependency change itself looks correct, the focused tests pass, and the current GitHub build is green.

danielbrownjr:
Implemented the remaining P2 in `9afad27`.

DragonBreath now derives the active integral ceiling as `approach_output_max - current_proportional_term` (bounded to the existing `[0, 1]` integral range) and passes it to upstream `dc_pid`. When the approach range contracts, `dc_pid` transactionally normalizes the stored integral to that tighter bound before stepping. This keeps stored P+I demand inside the actuator range while preserving derivative/measurement history and the existing `integrate=false` safety-inhibition path. There is no second PID implementation, post-controller masking fix, or controller reset.

The new regression uses one continuous controller state through the real trajectory, without resets:

- 100% steady: I `0.700000`, P+I/raw/duty `1.000000`
- 100% -> 70%: I `0.500000`, P+I `0.700000`, retained-D raw/duty `0.636000`
- 70% -> 40%: I `0.300000`, P+I `0.400000`, retained-D raw/duty `0.284800`
- after 4,000 additional 40% samples: I `0.300000`, P+I/raw/duty `0.400000`
- 40% -> 70% relaxation: I changes by one normal `0.000900` increment; raw/duty `0.482900`

This covers both downward transitions, sustained 40% operation, bounded stored demand, absence of a negative kick, and upward relaxation. Existing fixed-cap, integration-hold/foldback, source, SSR, and heater-policy tests remain passing.

P3 remains fixed: the host runner fetches `justinh-rahb/dragon-core` tag `v0.31.0`, matching the production manifest. Local validation passed all nine C host runners, 14/14 HIL runner tests, API v2 contract checks, a clean ESP32-C3 production build with warnings as errors, both safe HIL builds, and both GPIO/ADC compile-out checks. No heating test was performed.

justinh-rahb (changes requested):
The two prior review findings are addressed on 9afad27.

- The continuous 100%→40% reproduction now normalizes integral from 0.7 to 0.3, with stored P+I and delivered duty both at 0.4.
- tests/run_heater_pid_host_test.sh now fetches upstream justinh-rahb/dragon-core v0.31.0, matching the firmware manifest.
- The focused host test and current GitHub build pass.

One hardware-backed test issue remains:

[P2] Update the physical AUTO HIL scenario for source-target policy — tests/hil/scenarios/devboard-auto.json

I built and flashed the GPIO-safe native-USB profile to an attached ESP32-C3 and ran the checked-in --suite. devboard-auto.json failed at “AUTO engages”: it injects connected=true and bed_c=100 but no src_target_c, while current pb_policy deliberately engages AUTO only when the connected source reports src_target_c > 0. The board correctly returned auto_engaged=false, so the stale scenario prevents the documented full physical HIL suite from passing.

On the same board, a corrected scenario supplying src_target_c=55 and chamber_src_c=25 passed the engage/disconnect flow over native USB. The existing button and safety/ownership scenarios also passed, and every response reported mains_gpio_compiled_out=true.

Please update devboard-auto.json to inject the current source target/chamber telemetry and assert the new policy. This should also make each scenario start from an explicit safe state so an earlier failure does not cascade into the next scenario.

danielbrownjr:
Thanks for the hardware-backed review. The requested HIL correction is complete on a79bddd.

- devboard-auto now supplies src_target_c=55 and chamber_src_c=25, and asserts AUTO engagement plus effective_target_c=55.
- Every checked-in HIL scenario now begins from an explicit OFF safe state, with HIL runner coverage preventing regression.
- The branch integrates upstream main 193e278 and uses upstream dragon-core v0.32.0; dc_pid is unchanged between v0.31.0 and v0.32.0.
- Validation is green: 9/9 C host runners, 16/16 HIL runner tests, API v2 contract, production ESP32-C3, native-USB and UART safe-HIL, production-network debug, and all GPIO/ADC compile-out checks. GitHub Actions also passes cppcheck and artifact upload.

No controller or safety behavior changed; this update is limited to HIL correctness/isolation and conservative upstream integration.

justinh-rahb (commented):
Real-Panda HIL on `a79bddd` surfaced a zero-cross issue that the current smoke scenario does not detect.

With the board forced OFF (`heater_output=false`, `fan_percent=0`), five one-second samples advanced `zero_cross_count` by roughly 206–217 edges/s. The reported last-edge interval alternated between about 0.98–0.99 ms and 7.34 ms. On 60 Hz mains we expect one usable edge per half-cycle (~8.33 ms), so this looks like a duplicate/noisy edge being accepted inside each half-cycle. `zcd_isr()` currently accepts every positive edge without minimum-spacing filtering, and fan turn-on is applied on any accepted edge; that could move the actual switch point roughly 1 ms away from the true crossing.

Please filter implausibly close ZCD edges (with a threshold compatible with both 50 and 60 Hz), and extend the Panda smoke qualification to validate cadence/interval rather than only checking that the count is nonzero. The separate bounded 45 C hot test itself passed: SSR output modulated ON → OFF → ON across consecutive 10 s windows, PTC rose 31.9 → 33.4 C, no fault latched, and the final plus independent cleanup both verified OFF.

justinh-rahb:
Long-run real-Panda follow-up on `a79bddd`: the PID/heater path passed extended supervised qualification.

- 3-minute cold soak at 45 C: chamber 32.4 -> 37.2 C, PTC 33.2 -> 69.5 C; SSR window modulation and every lease renewal passed; no fault.
- 10-minute continuation: chamber 37.2 -> 42.9 C, PTC settled near 58.7 C; the run exercised the 70% and 40% approach regions with no latch, inhibition, or lease loss.
- 15-minute hot-start/hold: the printer bed independently ramped 39 -> 100 C, so this segment is not a Panda-only capacity measurement. Chamber reached 45 C at about 450 s. Panda output remained OFF at/above target for the rest of the run while external bed heat carried chamber temperature to 46.8 C; there was no integral-driven overshoot or fighting of external heat. A brief window was correctly issued when chamber dipped to 44.9 C before the sustained crossing.
- Scenario-final OFF and a separate non-heating smoke cleanup both passed. Final state was mode OFF, heater output OFF, no fault.

I did not find a further PID/control defect in the extended hot test. The separately reported zero-cross duplicate-edge finding remains the open hardware issue.

justinh-rahb (commented):
Real-hardware comparison found a DragonBreath product-tuning regression, separate from the correctness of `dc_pid` itself.

Using release `v1.1.15` through the production Klipper helper with the bed at 95–100 C, the old bang-bang controller drove continuously to target: chamber crossed 45 C at about 90 s, PTC peaked around 82.7 C, chamber peaked around 45.9 C, then output stayed off while the 100 C bed held the enclosure near/above target. No fault occurred.

The PR wrapper is much more conservative than that established behavior. With `Kp=0.025`, a normal 10 C warm-up error contributes only 25% proportional duty. `Ki=0.0003` adds only about 0.003 duty/s at that error, so it takes minutes to build toward the nominal 70%/100% approach ceilings. The 70% ceiling begins 10 C away and the 40% ceiling begins 5 C away, further slowing most of a typical 35 -> 45 C warm-up. This matches the real-Panda PR runs, where element temperature and chamber rise were substantially tamer/slower than release.

The generic controller, cap-aware anti-windup, source reset, and at/above-target behavior all looked correct in hardware. The issue is the DragonBreath-specific constants/policy. A reasonable first hardware sweep would be:

- raise `Kp` from 0.025 to approximately 0.10 so the controller immediately reaches the intended caps at ordinary warm-up errors;
- retain `Kd=0.04` initially;
- compare the current `Ki=0.0003` against roughly 0.001 for faster steady-demand buildup; and
- if closer release warm-up parity is desired, move the 70%/40% transitions closer to target (for example 5 C / 2 C rather than 10 C / 5 C).

Those are starting sweep values, not claimed final tuning. Please tune and repeat a matched real-Panda A/B with the bed held at 100 C and comparable starting chamber temperature. Host fixed-error tests prove bounds and anti-windup, but cannot establish acceptable thermal response for this heater.

justinh-rahb:
I reran the hot HIL comparison with the bed at operating temperature and tested a more aggressive local tuning candidate on the current PR head (`a79bddd`). The generic `dc_pid` primitive still looks sound; the behavior difference is attributable to the DragonBreath wrapper constants/caps.

Candidate A (local experimental build, not a PR commit):

```c
#define PB_HEATER_PID_KP 0.100f
#define PB_HEATER_PID_KI 0.001f
#define PB_HEATER_PID_KD 0.040f

// output cap by positive error
error < 2 C: 40%
error < 5 C: 70%
otherwise: 100%
```

Matched 45 C chamber test with the bed at 95-100 C:

| Firmware | Time to first reach 45 C | PTC peak | Initial chamber overshoot |
|---|---:|---:|---:|
| Release v1.1.15 | about 90 s | 82.7 C | 45.9 C |
| Candidate A | 116.6 s | 66.4 C | 45.2 C |

Candidate A passed the full 10-minute heater-enabled Panda HIL scenario, including lease renewal and safety assertions. Once chamber temperature exceeded the target, the Panda heater output remained off; the later climb to about 47.2 C was driven by the 100 C bed rather than controller demand. Final OFF cleanup and an independent smoke/OFF check also passed.

This seems like a substantially better DragonBreath starting point than the constants currently in the PR: only about 27 s slower than the old bang-bang controller in this run, while reducing element peak temperature by roughly 16 C and immediate overshoot by roughly 0.7 C. I recommend adopting these values (and updating the host expectations), then rerunning CI and HIL. The zero-cross duplicate-edge review item remains separate from this tuning result.

plastikman:
## Hardware A/B on a real Panda — DragonBreath PID tuning results

Following up @justinh-rahb's tuning comment with a matched real-Panda A/B. This walks through every step so the tuning decisions are reproducible.

### Setup
- **Build:** `hil-panda` profile of this PR's controller (`dc_pid`) flashed to a real Panda Breath (ESP32-C3, mains SSR + PTC element live), driven over the serial HIL console (`power_on` / `state`), polling chamber + element every ~5 s with a `PTC>103 °C` external abort and always-off-on-exit.
- **Thermal environment:** bare bench — **no printer bed, no enclosure, no aux fan.** This is the worst case for coupling element heat into the chamber, so absolute temps are lower than a real print; the value here is the *relative* controller comparison.
- **Safety:** every run relied on the firmware's element foldback + 105 °C hard trip; the external abort never had to fire.

### The tuning journey (soak to 60 °C, warm start)

| Config | Peak chamber | Peak element | Notes |
|---|---:|---:|---|
| PID `Kp=0.025, Ki=0.0003`, bands 10/5 (**original**) | 50.6 °C | 93 °C | slow; stalled ~50 °C |
| PID `Kp=0.10, Ki=0.001`, bands 10/5 | 55.1 °C | 94 °C | better, but throttles at ~55 °C (70 % cap at a 5 °C error) |
| **v1.1.15 bang-bang** (release baseline) | 58.7 °C | 101.5 °C | drives 100 % to foldback |
| **PID `Kp=0.10, Ki=0.001`, bands 5/2** (Justin's exact sweep) | 58.4 °C | 97.6 °C → **settles 76 °C** | matches bang-bang's reach; element *cooler* once holding, no sustained foldback |

**Findings, on hardware:**
1. **The original constants were the "slow" problem** — confirmed. `Kp=0.025` gives only ~25 % duty at a 10 °C error and the 10 °C/5 °C bands throttle most of the warm-up, so the chamber crawls and stalls ~50 °C.
2. **The approach-band placement was the real cap.** At `Kp=0.10` with 10/5 bands it still plateaued at 55 °C (70 % cap kicks in 10 °C out). Tightening to **5 °C/2 °C** lets it hold 100 % down to a 5 °C error and it reaches ~58 °C — matching the release bang-bang.
3. **The tuned PID is gentler than bang-bang for the same result.** Bang-bang parks the element at the 101 °C foldback to reach 58.7 °C; the tuned PID reaches 58 °C and then *settles the element to ~76 °C* while holding, never sitting on the foldback. Same chamber temp, cooler element, stable regulation instead of foldback cycling.

### 30 → 70 °C ramp (Justin's exact config: `Kp=0.10, Ki=0.001, Kd=0.040`, bands 5/2)
From a ~36 °C cold start: **~60 °C reached at ~20 min**, asymptoting to **62.4 °C** with the element foldback-limited at 102 °C. It cannot reach 70 °C on this bench — the element is already maxed at foldback, so ~62 °C is the heat-in/heat-out equilibrium here. **70 °C needs the printer bed + aux fan** (external enclosure heat + airflow to couple the element into the chamber) — consistent with the known "no airflow → hot element, cool chamber" behavior.

### Safety validation
The firmware element foldback engaged exactly as designed on both the bang-bang and PID paths — cutting the SSR at ~101–102 °C to hold the element under the 105 °C hard trip. No fault latched, and the external `PTC>103` guard never triggered. Product-owned safety layers stayed authoritative throughout.

### One observation for testers
When the device is linked to Klipper/Moonraker and the printer is idle, the `klipper` source will assert "no heat" and yank a manual `power_on` (seen once mid-run: `source` flipped `web → klipper`, `mode → off`). Disabling the source cleared it. Not a bug — just source arbitration to be aware of when bench-testing.

### Recommendation
This validates @justinh-rahb's suggested sweep on real hardware. Suggest updating the PR's constants to:
```c
#define PB_HEATER_PID_KP  0.100f   // was 0.025
#define PB_HEATER_PID_KI  0.001f   // was 0.0003
#define PB_HEATER_PID_KD  0.040f   // unchanged
// approach caps: error<2 → 40%, error<5 → 70%, else 100%  (was <5 → 40%, <10 → 70%)
```
The `dc_pid` port itself — generic controller, cap-aware anti-windup, source-transition reset, at/above-target behavior, and foldback ownership — all behaved correctly on hardware. The remaining item is these DragonBreath-specific constants, and this A/B supports the values above.

<sub>🤖 Hardware A/B via Claude Code.</sub>

danielbrownjr:
Physical Panda ZCD cadence qualification is complete on the exact current PR head `eec473f9a0e1929ea72715b621f98e44f58e47cb`.

I built the Panda HIL profile locally, flashed it to a real Panda, and ran the checked-in non-heating smoke scenario over the CH340K UART bridge:

`python tools\hil.py --target panda --port COM3 --scenario tests\hil\scenarios\panda-smoke.json --verbose`

Result: **PASS**

- Accepted zero-cross count increased by 609 over approximately five seconds: approximately **121.8 accepted edges/s**.
- Accepted intervals remained near the expected 60 Hz half-cycle: **8,316 us initially, 8,355 us after the observation interval, and 8,313 us finally**.
- The rejected-edge counter increased by 325, confirming the duplicate/noisy input edges were present and rejected by the new 4,000 us minimum-spacing filter.
- No accepted approximately 1 ms duplicate cadence was observed.
- Heater remained OFF, fan remained 0%, mode remained OFF, and no fault latched.

Local evidence is retained in `hil-results\20260904-070546\report.json` and `hil-results\20260904-070546\serial.jsonl`.

This closes the previously outstanding real-Panda non-heating ZCD cadence qualification gate. @justinh-rahb, please perform the final review when convenient.

justinh-rahb (approved):
Approved at `eec473f9`.

The DragonBreath-specific tuning now matches the hardware-validated configuration (`Kp=0.100`, `Ki=0.001`, `Kd=0.040`, with 40%/70% approach caps at 2 C/5 C). The cap-aware anti-windup, transition normalization, HIL scenario corrections, upstream `dc_pid` dependency, and physical Panda hot-test results are satisfactory.

The duplicate zero-cross edge issue is also addressed with minimum-spacing filtering and has now passed physical-Panda cadence qualification: accepted edges track the expected 60 Hz half-cycle cadence while close duplicate/noisy edges are rejected. CI is green. No remaining review blockers from me.

PR #92, opened 2026-08-31
ui: prototype explanatory control diagnostics

danielbrownjr:
## Summary

Refreshes the downstream Dragon-family consequential-toggle prototype and sparse DragonBreath control diagnostics onto current main, including the merged shared-`dc_pid` chamber controller and PID duty telemetry.

## Diagnostics hierarchy

The primary `/diag` view now explains, in order:

- active controller and effective temperature source
- process variable and target
- normalized controller request
- output allowed after DragonBreath limits
- actual instantaneous SSR state
- dominant constraint and overall safety/fault state

Lower-priority sensor/source details remain collapsed. The trend remains limited to process variable, target, PTC temperature, and allowed heater output.

## Firmware/API semantics

Adds a read-only, controller-agnostic `control.loop` snapshot to `/api/v2/state`:

- `controller` reports the active `pid` implementation.
- `preferred_source` reports the persisted Bambu chamber-source preference.
- `effective_source` and `process_variable_c` are captured from the exact source/value supplied to the real `dc_pid` step; they are unavailable while the controller is not running.
- `controller_request` is the normalized P+I+D request reconstructed from the authoritative `dc_pid_result_t` terms before the active approach ceiling and local thermal governors. It is observational only and is never fed back into control.
- `allowed_output` reuses the existing authoritative post-approach/post-thermal-governor `heater.commanded_duty`.
- `heater.output` remains the actual instantaneous SSR command.
- `constraint` reports the dominant product-owned approach, target, foldback, controller-error, or safety-inhibit reason.

No shadow PID or duplicate actuator calculation is introduced. Serialization remains read-only and no mutation endpoint is added.

## Consequential-toggle prototype

- Keeps a product-local, framework-free toggle/modal component.
- Uses the existing Automatic filtration setting as the truthful integration point.
- Enabling explains the selected printer-source/bed-target consequence and requires explicit confirmation.
- Cancel performs zero mutations; Confirm performs exactly one existing authenticated product settings mutation.
- Pending, disabled, failure, duplicate-submit, keyboard, mobile, and authoritative-server reconciliation states are covered.
- Product policy and safety remain firmware-owned.

## Behavior and safety scope

This PR does not change PID gains or math, approach thresholds, local/PTC foldbacks, hard faults, interlocks, Bambu eligibility/fallback policy, target policy, fan behavior, SSR timing, SSE lifecycle, or HTTP memory behavior. The controller still consumes the same process variable and drives the same governed duty path.

## Validation

Passed after merging current main:

- 10/10 C host-test runners
- consequential-toggle UI tests: 12/12
- control-diagnostics UI tests: 8/8
- HIL runner tests: 19/19
- API v2 static contract
- JavaScript syntax checks
- production ESP32-C3 build with first-party warnings as errors
- native-USB safe-HIL build
- UART safe-HIL build
- production-network debug build
- all GPIO/ADC and HIL compile-out checks
- `git diff --check`
- rendered desktop and exact 390 px mobile checks, including the opaque modal backdrop, visible cancel focus, stacked mobile actions, and zero horizontal overflow

No new physical-HIL run was performed for this UI refresh. The unchanged merged PID/SSR path already had real-Panda partial-duty evidence on the telemetry base from PR #97 (`a5a9253`): commanded duty varied from 29.7% to 55.0% during heated stress, with no fault or inhibit. The new request-vs-allowed distinction also has host regression coverage where the controller request exceeds the active 40% approach ceiling while actuator-facing duty remains capped.

## Future `dc_ui` candidate

`docs/consequential-toggle-ui.md` records the smallest plausible presentation-only API. Nothing is upstreamed to dragon-core or `dc_ui` here. A second Dragon product, preferably Jump Jet, should validate the shape before generalization; transition eligibility, side effects, persistence, restoration, and safety validation remain product-owned.

## Deliberately not included

- no private PID backend or controller selector
- no controller or safety behavior change
- no new Bambu chamber-control behavior
- no dragon-core/`dc_ui` change
- no SSE lifecycle work
- no HTTP memory work


plastikman:
## Code review — 5 findings

_(Automated review, high effort; verified against the branch. Line refs corrected — the `db_portal.c` embed helper is at ~71–77, not 183–185.)_

### 1. Served JS carries a stray trailing NUL — `components/db_portal/db_portal.c:77`
`embedded_text()` does `httpd_resp_send(req, start, end - start)`, but ESP-IDF's `TEXT` embed appends a `\0` before `_binary_..._end` (confirmed in `data_file_embed_asm.cmake`), so `end - start == filesize + 1`. `/diag` HTML and both `/ui/*.js` responses (consequential-toggle.js, control-diagnostics.js) get a trailing NUL and a `Content-Length` one byte too long. HTML tolerates it, but a NUL in served `text/javascript` is malformed and can trip strict parsers. **Fix:** `end - start - 1` (or `strlen`) for TEXT embeds. New regression vs. the old chunked `diag_page`.

### 2. The "requested vs allowed" reading is largely illusory — `components/pb_heater/include/pb_heater_pid.h` (`controller_request`)
`controller_request` (`result.p+result.i+result.d`) is presented as demand *before* the approach ceiling, but `dc_pid` already clamps the integral to `output_max - proportional` (with `output_max` = the approach cap) upstream. In real heat-up the reported request ≈ `allowed_output`; they only diverge on a one-tick derivative transient (the sole case the host test exercises), and `percent()`'s `Math.round(value*100)` collapses them further. The headline request-vs-allowed distinction barely shows a difference in normal operation — worth reconsidering what this reading is meant to convey before building UI around it.

### 3. `/diag` ignores the user's forced theme — `components/db_portal/www/diagnostics.html`
The rewrite dropped the `db_theme` localStorage restore + `[data-theme=light]/[data-theme=dark]` CSS the old page had; it now only uses `light-dark()` gated on `prefers-color-scheme`. A user who forced dark/light in the SPA gets a `/diag` that ignores it — theme-inconsistent with every other page. Restore the `db_theme` read + `data-theme` application.

### 4. First SSE blip → permanent 2 s polling — `components/db_portal/www/control-diagnostics.js` (`connect()` `onerror`)
`stream.onerror` calls `stream.close()` and falls back to `fetch('/api/v2/state')` polling on the **first** error — even a transient `CONNECTING`-state hiccup EventSource would auto-reconnect. Once tripped, live 1 Hz telemetry stays at 2 s polling until page reload. Prefer letting EventSource auto-reconnect (only fall back on repeated failures, or gate on `readyState === CLOSED`).

### 5. `no-store` on immutable JS — `components/db_portal/db_portal.c:76`
`embedded_text()` sets `Cache-Control: no-store` for all assets, but the versioned `/ui/*.js` change only with a firmware image. That forces a full re-download of both scripts on every `/diag` visit over the already-constrained single-worker httpd. Keep `no-store` for the live-telemetry HTML, but give the static JS `immutable`/`max-age` (or an ETag).

---
**Priority:** #1 is a clear one-line correctness fix; #2 is a design question (does the feature deliver its headline distinction?) worth settling before UI polish; #3–#5 are quality/UX regressions off the old `/diag`.

components/db_portal/include/db_portal_assets.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "esp_http_server.h"
+
+#define DB_PORTAL_HTML_CACHE "no-store"
+// URLs are stable across OTA: bound freshness rather than promising immutable
+// content for a year. Reload after 60 s to guarantee the new firmware's JS.
+#define DB_PORTAL_JS_CACHE "public, max-age=60, must-revalidate"
```
plastikman, line 8: **Cache skew after OTA (medium).** JS assets get `max-age=60, must-revalidate` on unversioned URLs while the HTML is `no-store`. After an OTA that changes both `diagnostics.html` and `control-diagnostics.js`, a visitor within the 60s window gets the fresh `no-store` HTML but the still-fresh cached old JS without revalidating (`must-revalidate` only applies once `max-age` expires) — the new HTML calls an API the old JS lacks and `/diag` breaks until the cache expires.

Fix: reuse the existing `dc_portal.c` `spa_get` pattern — `Cache-Control: no-cache` + an ETag from `esp_app_get_description()->app_elf_sha256`, giving a cheap 304 when unchanged and skew-free pickup after OTA.

components/db_portal/www/diagnostics.html
```diff
@@ -0,0 +1,103 @@
...
+  window.addEventListener('DOMContentLoaded',function(){
+    DBConsequentialToggle.mount($('filter-toggle'),{
+      id:'automatic-filtration',label:'Automatic filtration',description:'Allow the chamber blower to follow the connected printer source\'s bed target while the heater is idle.',pendingMessage:'Saving automatic filtration…',successMessage:'Automatic filtration state reconciled with DragonBreath.',
+      read:function(){return json('/settings',{cache:'no-store'}).then(function(s){return filterState(s.filter_auto);});},
+      confirm:function(requested,state){if(!requested)return null;return{severity:'warning',title:'Enable automatic filtration?',body:'DragonBreath will run the chamber blower whenever the selected printer source is connected and its bed target reaches the saved filtration threshold, even while the heater is idle.\n\nThe saved threshold is retained if you turn automatic filtration off later.',cancelLabel:'Cancel',confirmLabel:'Enable filtration'};},
+      mutate:function(requested){return json('/settings?filter_auto='+(requested?1:0),{method:'POST',headers:{'X-DragonBreath-Auth':auth()}}).then(function(s){return filterState(s.filter_auto);});}
```
plastikman, line 93: **Toggle has no 403 recovery (medium).** With a control token configured and no cached `db_tok` (user opens `/diag` before authenticating via the main dashboard), `auth()` returns the `web` sentinel; enabling the toggle POSTs `/settings`, `auth_reject` returns 403, `json()` throws, and the toggle just reverts to off with no way to enter the token. The main SPA prompts for the token and retries on 403; `/diag` offers no such path, so filtration can never be enabled from this page.

components/db_portal/www/diagnostics.html
```diff
@@ -0,0 +1,103 @@
...
...te(s[k]))maxTemp=Math.max(maxTemp,s[k]);});});maxTemp=Math.ceil(maxTemp/10)*10;x.font='10px sans-serif';x.fillStyle=colors.grid;x.strokeStyle=colors.grid;x.lineWidth=.5;x.textBaseline='middle';[0,.5,1].forEach(function(f){var y=top+(1-f)*plotH;x.beginPath();x.moveTo(left,y);x.lineTo(left+plotW,y);x.stroke();x.textAlign='right';x.fillText(Math.round(maxTemp*f)+'°C',left-4,y);x.textAlign='left';x.fillText(Math.round(f*100)+'%',left+plotW+4,y);});function px(i){return left+i*plotW/(samples.length-1);}function py(v){return top+(1-v/maxTemp)*plotH;}function line(key,color){x.beginPath();x.strokeStyle=color;x.lineWidth=2;x.setLineDash([]);var begun=false;samples.forEach(function(s,i){var v=s[key];if(!finite(v)){begun=false;return;}if(!begun){x.moveTo(px(i),py(v));begun=true;}else{x.lineTo(px(i),py(v));}});x.stroke();}line('pv',colors.pv);line('target',colors.target);line('ptc',colors.ptc);x.beginPath();x.strokeStyle=colors.duty;x.lineWidth=1.5;x.setLineDash([5,3]);var dutyBegun=false;samples.forEach(function(s,i){if(!finite(s.duty)){dutyBegun=false;return;}var y=top+(1-s.duty)*plotH;if(!dutyBegun){x.moveTo(px(i),y);dutyBegun=true;}else{x.lineTo(px(i),y);}});x.stroke();x.setLineDash([]);}
```
plastikman, line 82: **Chart theme fallback (low).** When `window.matchMedia` is unavailable (or stubbed) and no explicit `db_theme` override is set, `dark = theme==='dark' || (theme!=='light' && window.matchMedia && ...)` evaluates false, so the trend canvas draws the light palette (dark grid/line) on a dark page background. The page CSS itself follows the system theme via `light-dark()`, so only the canvas mismatches.

components/pb_heater/pb_heater.c
```diff
@@ -405,9 +437,13 @@ static void do_latch(pb_fault_reason_t code, const char *reason,
...
     s_fault_code = code;
     s_fault_reason = reason ? reason : pb_heater_fault_str(code);
+    s_effective_external = false;
+    s_process_variable_valid = false;
+    s_process_variable_c = 0.0f;
+    s_pid_duty = 0.0f;
     s_commanded_duty = 0.0f;
     s_approach_limit = 0.0f;
-    s_constraint = PB_HEATER_CONSTRAINT_OFF;
+    s_constraint = PB_HEATER_CONSTRAINT_SAFETY_INHIBITED;
```
plastikman, line 446: **New enum value the pinned SPA may not know (medium).** `do_latch` now sets the pre-existing public `heater.constraint` field to `safety_inhibited` where it previously reported `off`. The dc_ui dashboard is pinned at v0.33.0 and unchanged in this PR, so if it maps constraint values to labels it will render the unfamiliar string raw or as 'unknown' during a fault. `api-v2.md` was updated but the consumer wasn't — worth confirming the pinned SPA tolerates the new value.

plastikman (commented):
Automated code review (high effort). Tests pass and the `control.loop`/`telemetry_set`/dc-core wiring checks out — no critical/high issues. 5 medium/low findings inline below.

components/pb_httpd/pb_httpd.c
```diff
@@ -228,6 +228,22 @@ static cJSON *state_json(const pb_policy_snapshot_t *s)
...
+    cJSON_AddStringToObject(loop, "preferred_source",
+                            heater_telemetry.preferred_external ? "bambu" : "local_ntc");
+    cJSON_AddStringToObject(loop, "effective_source",
+        !heater_telemetry.process_variable_valid ? "unavailable" :
+        heater_telemetry.effective_external ? "bambu" : "local_ntc");
+    if (heater_telemetry.process_variable_valid)
+        add_num1(loop, "process_variable_c", heater_telemetry.process_variable_c);
+    else
+        cJSON_AddNullToObject(loop, "process_variable_c");
+    add_num3(loop, "pid_output", heater_telemetry.pid_duty);
```
plastikman, line 242: **Missing isfinite guard (low).** `add_num3` (used for the new `pid_output`/`allowed_output`) has no non-finite guard, unlike `add_num1` which emits `null`. If `pid_duty`/`commanded_duty` were ever NaN/Inf, `snprintf("%.3f", nan)` + `cJSON_AddRawToObject` emits bare `"pid_output":nan`, invalid JSON that breaks `JSON.parse` for every `/api/v2/state` and SSE client. Low likelihood (duty is clamped to finite [0,1]), but this PR routes a new value through the unguarded helper; mirror `add_num1`'s non-finite→null guard.

Issue #95, opened 2026-09-04
PID approach limits cause prolonged oscillation and persistent ~2 °C setpoint offset


danielbrownjr:
## Summary

A supervised bare-metal soak test of the PID implementation merged in #91 shows bounded and safe behavior, but it does not regulate acceptably at a 55 °C manual target. After the initial transient, the controller remains materially below setpoint and cycles across the 40%/70% approach-limit boundary.

This should block the first general release containing the PID controller. The latest tagged release, v1.1.15, predates #91 and is not affected.

## Test configuration

- Hardware: Panda Breath V1.0 / 33 kΩ Rref variant
- Firmware: `5e2f668` (merge of #91)
- Mode: Manual / POWER_ON
- Target: 55 °C
- Capture: DragonSniff 15-minute Soak profile
- State interval: 2 seconds
- Health interval: 30 seconds
- Test supervised throughout

## Results

| Metric | Result |
|---|---:|
| Initial temperature | 27.7 °C |
| First setpoint crossing | 54 seconds after enabling heat |
| Initial peak | 55.9 °C |
| Final five-minute range | 50.5–55.1 °C |
| Final five-minute mean | 52.64 °C |
| Final five-minute peak-to-peak swing | 4.6 °C |
| Samples within ±1 °C | 24% |
| Approximate delivered output | ~60% |

The initial overshoot is modest, but the controller never settles near the requested target. The observed cycle is consistent with the current product policy:

- Error >= 5 °C: up to 100%
- Error 2–5 °C: capped at 70%
- Error < 2 °C: capped at 40%
- At/above target: commanded off

The hardware appears to require roughly 60% average output to hold 55 °C. The 40% near-target ceiling therefore cannot maintain equilibrium: temperature falls below 53 °C, the ceiling rises to 70%, temperature recovers, and the cycle repeats.

## Safety and resource observations

- No fault or inhibit
- No reboot or boot-ID change
- No failed HTTP requests
- Minimum-free-heap watermark unchanged
- Heap remained stable after the dashboard SSE client connected
- No evidence of runaway heating or a resource leak

## Expected behavior

After reaching a stable thermal condition, chamber temperature should remain reasonably centered on the requested target without a persistent multi-degree limit cycle caused by the approach policy.

## Suggested next steps

1. Treat this as a release blocker for the next release containing PID control.
2. Rework the approach policy rather than only changing Kp/Ki; a near-target ceiling below required equilibrium duty structurally prevents setpoint tracking.
[...]

PR #97, opened 2026-09-04
feat(heater): expose PID duty telemetry

### LINKED ISSUE #96: Expose PID duty and active heater constraint in API v2
## Summary

Add read-only product-level heater telemetry to `GET /api/v2/state` so bare-metal validation can distinguish controller demand, approach limiting, downstream thermal inhibition, and the instantaneous SSR output.

This belongs in DragonBreath rather than dragon-core. `dc_pid` already exposes its generic output, terms, and saturation status; DragonBreath owns the approach policy, local/element foldback, safety authority, 10-second SSR window, and public product API.

## Motivation

The supervised 55 °C soak documented in #95 exposed a persistent multi-degree limit cycle. The existing API reports only:

```json
"heater": {
  "demand": true,
  "output": false
}
```

That is enough to observe an instantaneous SSR state, but not enough to identify the commanded duty or which product policy is constraining it. Sampling the boolean output can only approximate duty and can alias the 10-second window.

## Proposed additive state fields

```json
"heater": {
  "demand": true,
  "output": false,
  "commanded_duty": 0.400,
  "approach_limit": 0.400,
  "constraint": "approach_limit"
}
```

Suggested semantics:

- `commanded_duty`: normalized duty admitted to the SSR time-proportioning window after the PID approach policy and non-latching thermal governors
- `approach_limit`: current DragonBreath-owned maximum duty selected from control error
- `output`: existing instantaneous physical SSR command
- `constraint`: dominant explanation such as `off`, `none`, `approach_limit`, `target_reached`, `local_foldback`, `element_foldback`, or `pid_error`

Raw P/I/D terms should remain limited to advanced diagnostics/export if later needed; the normal state should explain actuator behavior without dumping controller internals.

## Requirements

- Additive API v2 change; preserve existing fields and behavior
- One lock-consistent telemetry snapshot
- Reading telemetry must not refresh a lease/watchdog or mutate PID state
- No change to control, approach, foldback, SSR, or safety behavior
- Document field semantics and stable constraint strings
- Add API contract coverage
- Verify host tests and an ESP32-C3 build
- Validate the resulting fields on hardware during the #95 reproduction/fix run

A behavior-neutral implementation is being prepared separately and should reference this issue.

danielbrownjr:
## Summary

- expose the duty admitted to DragonBreath's 10-second SSR window
- expose the active product-owned approach limit and dominant constraint
- keep the telemetry snapshot lock-consistent and observational only
- document the additive API v2 fields and constraint vocabulary
- add static API contract coverage and changelog entry

## Architecture

This remains DragonBreath-owned rather than moving into dragon-core. `dc_pid` supplies generic PID math and results; DragonBreath owns approach shaping, local and element foldback, SSR time proportioning, safety authority, and the product API.

The new state shape is:

```json
"heater": {
  "demand": true,
  "output": false,
  "commanded_duty": 0.400,
  "approach_limit": 0.400,
  "constraint": "approach_limit"
}
```

`commanded_duty` is post-approach and post-foldback: the normalized command actually admitted to the SSR window. `output` remains the instantaneous physical SSR command.

## Safety

This is behavior-neutral observability. It does not alter PID gains, approach thresholds, safety limits, fault handling, watchdog behavior, leases, or output timing. Reading the telemetry does not mutate controller state or refresh any authority.

## Validation

Passed locally:

- all `tests/run_*_host_test.sh` suites
- `tests/check_api_v2_contract.sh`
- `git diff --check`

ESP32-C3 firmware build remains for CI.

Closes #96.
Supports diagnosis and retest of #95.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ below into the GitHub Release notes.
 
 ## [Unreleased]
 
+### Added
+- Expose product-owned heater control telemetry in API v2 state snapshots:
+  commanded SSR-window duty, the active PID approach limit, and the dominant
+  controller or thermal constraint. This is read-only observability and does
+  not change heater control or safety behavior.
+
 ## [1.1.15] - 2026-09-01
 
 ### Fixed
--- a/components/pb_heater/include/pb_heater.h
+++ b/components/pb_heater/include/pb_heater.h
@@ -32,6 +32,28 @@ typedef enum {
                                   // corrupt and mapped to a generic latched fault
 } pb_fault_reason_t;
 
+// Read-only control telemetry for diagnostics and hardware validation. These
+// values describe DragonBreath's product-owned control path; dc_pid remains a
+// board-neutral math primitive with no knowledge of approach limits, safety
+// governors, or the physical SSR.
+typedef enum {
+    PB_HEATER_CONSTRAINT_OFF = 0,
+    PB_HEATER_CONSTRAINT_NONE,
+    PB_HEATER_CONSTRAINT_APPROACH_LIMIT,
+    PB_HEATER_CONSTRAINT_TARGET_REACHED,
+    PB_HEATER_CONSTRAINT_LOCAL_FOLDBACK,
+    PB_HEATER_CONSTRAINT_ELEMENT_FOLDBACK,
+    PB_HEATER_CONSTRAINT_PID_ERROR,
+} pb_heater_constraint_t;
+
+typedef struct {
+    // PID output after DragonBreath's active approach limit (normalized 0..1).
+    float commanded_duty;
+    // Product-owned maximum duty for the current temperature error (0..1).
+    float approach_limit;
+    pb_heater_constraint_t constraint;
+} pb_heater_telemetry_t;
+
 // Pure fail-safe decision for the boot-time fault restore (pb_heater_load_fault),
 // inline so it can be host-tested without an NVS backend. Given the outcome of
 // reading the persisted latch/code, decides whether to come up latched and with
@@ -213,6 +235,11 @@ esp_err_t pb_heater_init(void);
 //                           HTTP 409. Heat is never queued behind a fault latch.
 esp_err_t pb_heater_set_target_c(float target_c);
 float pb_heater_get_target_c(void);
+
+// Copy one lock-consistent diagnostic snapshot. This is observational only and
+// has no effect on PID state, watchdogs, leases, or heater output.
+void pb_heater_get_telemetry(pb_heater_telemetry_t *out);
+const char *pb_heater_constraint_str(pb_heater_constraint_t constraint);
 
 // Optional external chamber measurement used ONLY for set-point regulation.
 // Pass NAN to fall back to the local chamber NTC. The local chamber NTC and PTC
--- a/components/pb_heater/pb_heater.c
+++ b/components/pb_heater/pb_heater.c
@@ -37,6 +37,9 @@ static bool        s_persist_pending;  // guarded by s_mux — a latch persist n
 static int64_t     s_persist_retry_us; // guarded by s_mux — last persist-retry timestamp
 static bool        s_on;            // written only by the control task; atomic read
 static pb_heater_pid_state_t s_pid; // control-task only — shared dc_pid state + SSR window
+static float       s_commanded_duty; // guarded by s_mux; PID output after approach limit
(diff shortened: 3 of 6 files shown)


plastikman (approved):
## Code review — PID duty telemetry (high effort)

Clean, well-scoped addition. The "behavior-neutral" claim is the thing to verify, since this isn't pure telemetry — it also refactors the drive path — and it holds up.

### Behavior-neutrality — verified ✅
The drive computation changed from
`drive = !safety_inhibited && window_on(&s_pid, duty, now)`
to
`drive = window_on(&s_pid, commanded_duty, now)` with `commanded_duty = safety_inhibited ? 0 : duty`.

The only case that differs is `safety_inhibited`: before, `window_on` was short-circuited; now it's called with `0`. But `pb_heater_pid_window_on` returns `false` at `duty <= 0.0f` **before touching any state** — so during inhibit it's a no-op with identical output *and* identical window state (no `window_start_us` mutation). Non-inhibited path is unchanged. So output timing is genuinely preserved.

### Telemetry accuracy — verified ✅
- `commanded_duty` really is post-approach/post-foldback: `pb_heater_pid_step` bounds `*duty` by `approach_max_duty(error_c)` internally and returns `0` at/above target, and `commanded_duty` is forced to `0` under safety inhibit.
- `approach_limit` is recomputed with `target - regulation_c`, the same error `pid_step` uses, so it matches the internal cap.
- Constraint precedence (pid_error → element/local foldback → target_reached → approach_limit) is sound, and the `duty >= approach_limit - 5e-4` epsilon with the `approach_limit < 1.0` guard correctly reports `approach_limit` only when the cap is actually binding (and `none` when the PID is controlling freely below it).

### Locking — correct
`s_commanded_duty/approach_limit/constraint` are consistently guarded by `s_mux`: `telemetry_set` and `pb_heater_get_telemetry` both critical-section, `do_latch` zeroes them within its existing lock, and `init` seeds them. No torn reads.

### Two minor notes (non-blocking)
1. **The new logic is untested.** `check_api_v2_contract.sh` only greps the source for the field *names* — the constraint classification and the approach-limit epsilon (the actual new behavior) have no test. A small host test over `(error, foldback, pid_ok) → constraint` would lock the precedence in, the way the fault-restore logic is already host-tested.
2. **`add_num3` emits raw `%.3f`.** A non-finite `commanded_duty`/`approach_limit` would serialize as `nan`/`inf` → invalid JSON. It's practically unreachable (the fault path forces `duty = 0`), and `add_num1` shares the pattern, but an `isfinite(v) ? v : 0` guard in the raw-number helpers would harden the state document.

Nice work — accurate, lock-consistent, and the ownership boundary (product-owned shaping vs. `dc_pid` math) is respected. LGTM.

<sub>🤖 Automated review via Claude Code (high effort).</sub>

danielbrownjr:
## Bare-metal validation

Validated commit `a5a9253` on the 33 kΩ Rref Panda Breath hardware after a successful ESP32-C3 build and OTA update. GitHub CI is also green.

### Telemetry lifecycle

- idle: `demand=false`, `output=false`, `commanded_duty=0`, `approach_limit=0`, `constraint=off`
- active PID control: all new fields remained present and coherent in both HTTP state snapshots and SSE telemetry
- shutdown: demand, physical output, commanded duty, and approach limit all returned to zero; constraint returned to `off`
- no stale nonzero telemetry was observed after shutdown

### 30-minute thermal capture at 55 °C

The final 10 minutes averaged 53.18 °C with a 52.9–53.6 °C range and 39.9% mean commanded duty. The telemetry showed the controller hunting across the 2 °C approach-band boundary: 91/121 samples used the 40% approach limit and 30/121 used 70%; `approach_limit` was reported as the dominant constraint in 25 samples.

This provides direct evidence for the refined diagnosis in #95: the initial larger oscillation is a prolonged transient, followed by a narrow but persistent approximately 1.8 °C negative setpoint offset.

### Heated SSE Stress churn

DragonSniff Stress was run while manual PID control remained active at 55 °C:

- 20/20 SSE connections succeeded
- 128 events parsed successfully
- zero rejected connections
- zero HTTP, transport, local-resource, or parse failures
- no reboot, fault, or inhibit
- all local workers and connection permits cleaned up
- final settled SSE-client count returned to zero
- free heap recovered from 83,524 to 83,896 bytes
- minimum-free-heap watermark did not decrease during the run

The new telemetry remained coherent throughout: commanded duty ranged from 29.7–55.0%, with 96 `none` and 32 `approach_limit` constraint samples.

Raw JSONL is intentionally not attached because it contains local network and device identifiers.

PR #98, opened 2026-09-06
build: require fixed ESP-IDF async HTTP server

danielbrownjr:
## Summary

Require an ESP-IDF release containing Espressif's asynchronous HTTP-request socket-ownership fix.

- Reject ESP-IDF older than v5.3.2 at configure time.
- Pin CI production and safe debug/HIL builds to v5.3.5.
- Document the minimum version and upstream fix.

## Root cause

ESP-IDF v5.3.1 could return an asynchronous HTTP session's socket to the main HTTP task. During DragonBreath SSE churn, the main task closed and reused that descriptor while the asynchronous SSE worker still held its copied request. A stale SSE chunk could therefore be written into a newly accepted ordinary HTTP response, producing DragonSniff's observed `BadStatusLine: 11\r\n`.

Espressif fixed the concurrency defect in `8b559ce6145474f01fcbb8763604e6b8769ba84e`, first included in v5.3.2.

## Validation

- Configure-time rejection verified with local ESP-IDF v5.3.1.
- GitHub workflow is fully green using v5.3.5:
  - all host tests
  - static analysis
  - ESP32-C3 production build
  - API v2 contract
  - safe dev-board HIL builds
  - native-USB production-network debug build
- Instrumented v5.3.5 hardware build passed:
  - 10 observation/capture/observation repetitions
  - 20/20 sequential SSE churn connections
  - 150/150 concurrent state/health requests
  - 101 SSE telemetry events
  - zero HTTP, transport, parse, or ownership-loss failures
  - zero `BadStatusLine`
  - SSE count recovered to its pre-run baseline
  - boot ID and heap remained stable

No DragonBreath controller, heater, fan, safety, source-selection, HTTP endpoint, or SSE architecture behavior changes are included.

Hardware evidence is tracked in danielbrownjr/DragonSniff#10.

### DIFF
--- a/.github/workflows/firmware-build.yml
+++ b/.github/workflows/firmware-build.yml
@@ -115,10 +115,10 @@ jobs:
       # there is no committed lock to drift. The ESP-IDF build step re-solves and
       # validates the managed-component dependencies against the pinned manifests
       # (components/*/idf_component.yml) on every run, which is the validation.
-      - name: Build (ESP-IDF v5.3)
+      - name: Build (ESP-IDF v5.3.5)
         uses: espressif/esp-idf-ci-action@9d38657f3d789ca759b2b37aaf5ceffbc42c4f0d # v1
         with:
-          esp_idf_version: v5.3
+          esp_idf_version: v5.3.5
           target: esp32c3
           path: '.'
 
@@ -126,10 +126,10 @@ jobs:
         run: |
           sh tests/check_api_v2_contract.sh
 
-      - name: Build safe dev-board HIL profile (ESP-IDF v5.3)
+      - name: Build safe dev-board HIL profile (ESP-IDF v5.3.5)
         uses: espressif/esp-idf-ci-action@9d38657f3d789ca759b2b37aaf5ceffbc42c4f0d # v1
         with:
-          esp_idf_version: v5.3
+          esp_idf_version: v5.3.5
           target: esp32c3
           path: '.'
           command: |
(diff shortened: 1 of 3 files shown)


plastikman (approved):
## Code review — require ESP-IDF ≥ v5.3.2 (high effort)

Well-motivated and well-scoped: a build-time floor + CI pin for the ESP-IDF async-HTTP socket-ownership fix (`8b559ce`, first in v5.3.2) that was producing the `BadStatusLine: 11\r\n` corruption under SSE churn. No behavior change — purely build config + docs.

### The version guard is correct — verified ✅
The one place a check like this goes wrong is version comparison, and this gets it right. I replayed the exact `MATCHES` + `VERSION_LESS` logic across the edge cases:

| IDF_VER | result |
|---|---|
| `v5.3.1`, `v5.3.1-dirty` | REJECT ✅ |
| `v5.2.9` | REJECT ✅ |
| `v5.3.2`, `v5.3.2-123-gabc` | allow ✅ |
| `v5.3.5-dirty` | allow ✅ |
| **`v5.3.10`** | **allow ✅** (numeric compare — a string compare would wrongly reject it) |
| `v5.4.0`, `v6.0.0` | allow ✅ |

`VERSION_LESS` does true semver ordering (so `5.3.10 > 5.3.2`), the regex strips the `v` prefix and the `-dirty`/git-describe suffixes, and `CMAKE_MATCH_1` is captured correctly (nothing clobbers it before the `set`). Placement after `project()` is right — `IDF_VER` is only available once the IDF build system is initialized.

CI pins are consistent (all three build steps → `v5.3.5`, step names updated), and the README floor (`v5.3.2`) / CI note (`v5.3.5`) line up.

### One minor nit (non-blocking)
The regex requires exactly three numeric components, so a bare two-component `IDF_VER` (e.g. a `v5.3` branch report or a `v6.0-dev` master checkout) hits the "Cannot determine ESP-IDF version" `FATAL_ERROR` rather than being accepted. It fails *closed*, which is defensible, and 3-component is what releases and branch-describes actually emit — so this is theoretical. If you want to be lenient, either relax the capture to `([0-9]+\.[0-9]+(\.[0-9]+)?)` or use the `IDF_VERSION_MAJOR/MINOR/PATCH` variables (which sidestep string parsing entirely). Not required.

Clean, correct, and the guard actually does what it claims. LGTM.

<sub>🤖 Automated review via Claude Code (high effort).</sub>

PR #104, opened 2026-09-17
Add configurable device hostname

rmonk:
# Problem
The DHCP/mDNS hostname was hardcoded to "dragonbreath", causing conflicts for anyone running more than one device on the same network. 

# Fix
Adds a "Hostname" field to the web portal's Device section (NVS-backed, defaults to "dragonbreath"), validated as a DNS-safe RFC 1123 label. The value is read at boot and passed to dc_wifi_set_identity(); changes take effect after a restart.

# Verification
I tested this on two Panda Breaths on the same network and the hostname change happened cleanly on the DHCP/DNS side for both.


plastikman:
Great feature, this belongs 1/2 in dragon-core and 1/2 in each of the sub components.  this has been on the list of "we need to fix this" for a while, so thanks for raising it.  I plan to get this merged into 1.1.19 when its final.

PR #107, opened 2026-09-20
feat(heater): selectable bang-bang vs PID chamber control (fixes #95)

### LINKED ISSUE #95: PID approach limits cause prolonged oscillation and persistent ~2 °C setpoint offset

plastikman:
## Why

The #91 PID controller caps duty at **40 % within 2 °C of target**, but the hardware needs **~60 %** to hold, so it structurally can't settle and stalls ~2 °C low. Reported by several users; the supervised soak in #95 shows mean **52.6–53.2 °C** at a 55 °C target, approach-limited at 40 % for most samples.

Rather than only retuning the PID, this makes the control **method selectable** so a known-good fallback ships now while the PID path stays testable side-by-side.

## What

- `pb_heater_method_t {BANGBANG, PID}`, NVS-persisted (`app_nvs` / `heat_method`), `pb_heater_get/set_method()`, **default `BANGBANG`** (the v1.1.15 hysteresis on/off drive). Unknown/corrupt NVS values fall back to bang-bang.
- The control tick branches on the method **after** the shared safety/foldback steps. Both methods keep every safety layer unchanged — over-temp/sensor/comms trips (`pb_heater_eval_trip`) and the local + element (PTC) foldback governors. The PID/PTC path (v1.1.16, #91) is byte-for-byte unchanged, just opt-in. A method change resets both paths' history so the newly selected one starts clean.
- `heater.method` added to `/api/v2/state`; select via **`GET`/`POST /api/v2/heater_method`** (auth-gated, applies live, no reboot). Deliberately a v2 route, **not** part of legacy `/settings`.
- `docs/api-v2.md` + the api-v2 static contract test updated (contract test passes).

## Default behavior change

New installs and existing units both default to **bang-bang** on upgrade, so the stuck-at-−2 °C users get the known-good hold automatically. PID/PTC becomes opt-in for A/B testing.

## Testing

```bash
curl http://<breath>/api/v2/heater_method                 # {"method":"bangbang"}
curl -X POST http://<breath>/api/v2/heater_method \
  -H 'X-DragonBreath-Auth: <tok>' -d '{"method":"pid"}'   # switch to PTC for A/B
```
Run the DragonSniff soak on each and compare mean error / peak-to-peak.

## Follow-ups

- **UI:** a "Heater control method" selector on the dc_ui settings card (alongside the filament-profile config) — separate dragon-core PR + pin bump.
- Targeting **v1.1.19-rc2** once the UI lands.

Closes #95.


plastikman:
### Live validation — bang-bang holds setpoint (3.4 h print on hardware)

Flashed `feat/selectable-heater-method` (bang-bang default) to a real DragonBreath (`dragonbreath-8fe51c`, 82 kΩ board, was on v1.1.19-rc1) and logged the heater across a full **3.4-hour print** at a **55 °C** manual chamber target.

**Chamber hold (target 55 °C):**
- Reached target in ~5 min, held for ~206 min.
- Mean **54.73 °C**, range 53.6–55.6 (peak-to-peak **2.0 °C**).
- **92 % of samples within ±1.0 °C, 100 % within ±1.5 °C.**
- SSR duty ~29 % (bed at 90 °C assisting).

**Element safety:** PTC mean 67.3 °C, **peak 94.6 °C** — 10.4 °C under the 105 °C trip. Element foldback never engaged; the on/off cycling kept it safe. Zero faults; only `none`/`target_reached` constraints.

**vs. the original #95 PID capture** (same 55 °C target):

| Metric | PID (#95) | Bang-bang |
|---|---|---|
| Mean | 52.6 °C (−2.4) | **54.73 °C (−0.27)** |
| Within ±1 °C | 24 % | **92 %** |
| Peak-to-peak | 4.6 °C | **2.0 °C** |
| Holds setpoint? | No | **Yes** |

Bang-bang holds essentially on setpoint where the PID approach cap stalled ~2 °C low. It's the **default** in #107; the PID/PTC method stays selectable (now carrying #103's rate-gated cap fix) for continued A/B testing via `POST /api/v2/heater_method`.

Also verified live: `GET`/`POST /api/v2/heater_method` round-trips (`bangbang`↔`pid`), invalid values rejected (400), NVS-persisted, `heater.method` in `/api/v2/state`.

